PR #13, opened 2021-03-31
Proxy readiness channel cleanup

shaneutt:
- fix: correct a timing issue that would cause a closed chan panic
- fix: no report needed if the context is "done"

### DIFF
--- a/pkg/kind/kong_proxy.go
+++ b/pkg/kind/kong_proxy.go
@@ -11,7 +11,6 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/tools/cache"
 )
@@ -32,6 +31,9 @@ const (
 
 	// proxyRequestTimeout indicates the default max time we'll wait for a deployed Kong proxy to start responding to HTTP requests.
 	proxyRequestTimeout = time.Minute * 3
+
+	// time to wait between GET requests
+	serviceInformerTickTime = time.Millisecond * 200
 )
 
 // -----------------------------------------------------------------------------
@@ -50,7 +52,7 @@ type ProxyReadinessEvent struct {
 
 // startProxyInformer provides a channel indicates when the proxy server is fully functional and accessible
 // by providing the *url.URL by which to access it. The channel will produce a nil value on failure.
-func (c *kongProxyCluster) startProxyInformer(ctx context.Context) (ready chan ProxyReadinessEvent) {
+func (c *kongProxyCluster) startProxyInformer(ctx context.Context, timeout time.Time) (ready chan ProxyReadinessEvent) {
 	ready = make(chan ProxyReadinessEvent)
 
 	// we need to wait for the Kong proxy deployment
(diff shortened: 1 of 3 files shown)


pkg/kind/kong_proxy.go
```diff
@@ -109,7 +97,17 @@ func (c *kongProxyCluster) startDeploymentInformer(ctx context.Context) (deploym
 	return
 }
 
-func (c *kongProxyCluster) startServiceInformer(ctx context.Context, nsn types.NamespacedName, ready chan ProxyReadinessEvent) {
+// startServiceInformer exposes the provided deployment as LoadBalancer type and reports when the service has been provisioned a load balancer IP/Host
+func (c *kongProxyCluster) startServiceInformer(ctx context.Context, d *appsv1.Deployment, ready chan ProxyReadinessEvent) {
+	// expose the deployment first, if this fails then all fails
+	_, err := c.exposeProxyDeployment(ctx, d)
+	if err != nil {
```
mflendrich, line 104:
```suggestion
	if _, err := err := c.exposeProxyDeployment(ctx, d); err != nil {
```

pkg/kind/kong_proxy.go
```diff
@@ -109,7 +97,17 @@ func (c *kongProxyCluster) startDeploymentInformer(ctx context.Context) (deploym
...
+		ready <- ProxyReadinessEvent{Err: err}
+		close(ready)
```
mflendrich, line 106: Might be better to `defer close(ready)` at the beginning of this function instead of `close(ready)` before every `return`.

Why?
- This way we're guaranteed not to overlook channel closing
- less code
- `defer` will work also if this fn ever `panic`s

mflendrich (approved):
optional stuff follows

pkg/kind/kong_proxy.go
shaneutt: The problem is that there are conditions where this function should return without closing, the explicit closes cover the conditions where we should close and be done.

pkg/kind/kong_proxy.go
shaneutt: :+1: 5f019f238394200fba56aa803a907bf3872580ec

pkg/kind/kong_proxy.go
shaneutt: Sounds good, will implement.

shaneutt: :+1: bbfeaba1e62ab5036a0e198710a4559dd1ab1070

pkg/kind/kong_proxy.go
```diff
@@ -109,50 +100,67 @@ func (c *kongProxyCluster) startDeploymentInformer(ctx context.Context) (deploym
...
-	factory := kubeinformers.NewSharedInformerFactory(c.client, proxyInformerResyncPeriod)
-	informer := factory.Core().V1().Services().Informer()
-	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
-		// this UpdateFunc is responsible for closing the ready channel once the proxy URL is ready (or if an error occurs)
-		UpdateFunc: func(old1, new1 interface{}) {
-			newService, ok := new1.(*corev1.Service)
-			if !ok {
-				err := fmt.Errorf("somehow got unexpected type instead of corev1.Service: %T", new1)
+// startServiceInformer exposes the provided deployment as LoadBalancer type and reports when the service has been provisioned a load balancer IP/Host
+func (c *kongProxyCluster) startServiceInformer(ctx context.Context, d *appsv1.Deployment, ready chan ProxyReadinessEvent, timeout time.Time) {
```
mflendrich, line 104: I have several comments about how this function deals with contexts and timeouts:
- It appears that idiomatically in Go ["timeouts" are of type `time.Duration`](https://pkg.go.dev/context#WithDeadline) and ["deadlines" are of type `time.Time`](https://pkg.go.dev/context#WithDeadline). Here a `timeout` is a `time.Time` which left me confused.
- This function accepts a `ctx` (and passes it on) unrelated to the deadline. The deadline does not affect blocked operations accepting a context, but it could, virtually for free.
- There's a pattern for ticks with a deadline:

    ```go
    for timeNow := time.Tick(serviceInformerTickTime) {
      // option 1
      if timeNow.After(deadline) {
        break
      }
      // option 2
      select {
      case <-ctx.Done():
        break
      }
    }
    ```

My recommendation would be to drop the deadline altogether, and instead use `ctx`'s capabilities to implement timeout in this function.

shaneutt: My only concern has to do with one big timeout for an entire test (via the context) vs. timeouts for specific portions which can be more visible and tunable... :thinking: 

But I think for now I'll take your recommendation and we'll test with that for a while :+1:

shaneutt: After looking things over, I want to hold on making changes right now since this is blocking another PR and is otherwise functional, I created a follow up item to clean this up before our first release version: https://github.com/Kong/kubernetes-testing-framework/issues/14

PR #15, opened 2021-04-01
Add KONG_STREAM_LISTEN utilities

shaneutt:
This PR adds some functionality that is helpful for `TCPIngress` and `UDPIngress` tests, in that testing these requires being able to update the ENV and container ports of the running proxy to expose the new ports needed for TCP/UDP.

This is a first iteration in service of https://github.com/Kong/kubernetes-ingress-controller/pull/1150 but it is incomplete and there are follow up issues noted in the comments.

- chore: move the overrideenv util to k8s pkg
- chore: make the default vars public
- feat: add kong stream listeners cluster utils

### DIFF
--- a/pkg/kind/kong_stream_listen.go
+++ b/pkg/kind/kong_stream_listen.go
@@ -1,0 +1,235 @@
+package kind
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	"github.com/kong/kubernetes-testing-framework/pkg/k8s"
+)
+
+// -----------------------------------------------------------------------------
+// Timeouts
+// -----------------------------------------------------------------------------
+
+// FIXME: in the future we'll use context timeouts instead see: https://github.com/Kong/kubernetes-testing-framework/issues/14
+
+const (
+	// default timeout tick interval
+	waitTick = time.Second * 1
+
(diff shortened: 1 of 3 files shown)


pkg/kind/kong_stream_listen.go
```diff
@@ -0,0 +1,232 @@
...
+				ing := svc.Status.LoadBalancer.Ingress[0]
+				if ip := ing.IP; ip != "" {
+					provisioned = true
+					break
+				}
+			}
+			time.Sleep(waitTick)
+		}
+		if !provisioned {
+			err = fmt.Errorf("load balancer service for deployment %s did not provision successfully within %s", name, serviceWait)
```
shaneutt, line 219:
```suggestion
			err = fmt.Errorf("load balancer service for deployment %s did not provision successfully within %s", name, serviceWait)
			return
```

shaneutt: no reason to wait here if the lb didn't provision.

pkg/kind/kong_stream_listen.go
```diff
@@ -0,0 +1,232 @@
...
+		}
+	} else { // if the deployment didn't become ready within the timeout period, bail out there's nothing more we can do.
+		err = fmt.Errorf("deployment not ready after %s", timeout)
+	}
+
+	// FIXME: there appears to be a race condition that can occur in kong upstream if the Admin API is restarted and then updated
+	//        too quickly. Effectively the API will return 200 OK but it's not actually ready to accept updates. Given a few seconds
+	//        this will settle so as a workaround we're sleeping here, but we need to dig deeper into this problem to actually fix it.
```
shaneutt, line 228:
```suggestion
	//        this will settle so as a workaround we're sleeping here, but we need to dig deeper into this problem to actually fix it.
	//        See: https://github.com/Kong/kubernetes-ingress-controller/issues/941
```

shaneutt:
gonna just pull this for now, I'm going to have to rework a couple things.

PR #26, opened 2021-05-13
[chore] some general cleanup

shaneutt:
This PR does some general documentation cleanup and re-organization of some of the code, but doesn't change any functionality.


README.md
```diff
@@ -1,18 +1,26 @@
-**WARNING**: this is currently an _experimental_ tool from [@kong/team-k8s](https://github.com/orgs/Kong/teams/team-k8s) don't use this yet unless you're aware of the context. Check in with us in [#kong on Kubernetes Slack][slack]!
+# Kubernetes Testing Framework for Kong
 
-[slack]:https://kubernetes.slack.com/messages/kong
-
-# Golang Kubernetes Testing Framework Prototype
+Integration testing framework for deploying and testing [Kong][kong] on [Kubernetes][k8s].
 
-Integration testing framework for the [Kong Kubernetes Ingress Controller (KIC)][kic].
+The original intention was to provide testing facilities for integration testing the [Kong Kubernetes Ingress Controller (KIC)][kic].
```
rainest, line 5: IMO this should remain as-is, or use a single "testing Kong and the Kong Kubernetes Ingress Controller (KIC) on Kubernetes" line. We haven't stopped using this for testing KIC.

We do use "Kong on Kubernetes" in marketing material to refer to a suite of products rather than the individual components that comprise it, but using that alone can be a bit confusing in engineering contexts where the distinction between those components. As this is a testing framework with an engineering audience I'd lean towards the individual bits.

Also TIL [my thinking "original intent" is the correct variant is apparently a 90s thing](https://books.google.com/ngrams/graph?content=original+intent%2Coriginal+intention&year_start=1800&year_end=2019&corpus=26&smoothing=3&direct_url=t1%3B%2Coriginal%20intent%3B%2Cc0%3B.t1%3B%2Coriginal%20intention%3B%2Cc0).

rainest (approved):

README.md
shaneutt: Sounds good, changed this :+1:

PR #27, opened 2021-05-20
feat: fake kong admin api

shaneutt:
This adds a fake kong admin api which includes a buffer for response mocking. This was originally intended to support unit testing in the Kong Kubernetes Ingress controller for functions and methods which required a valid *kong.Client in order to be tested.

NOTE: this is not feature complete nor is it intended to be. For my immediate needs only the db configuration and version in the root config was needed, and the ability to override basic API requests. We can add more functionality to this fake admin api implementation later.


rainest (approved):

hbagdi:
My 2 cents:
We shouldn't try to mock Kong's Admin API with KIC and rather setup Kong as part of the testing infra.
One of the most important thing to test with KIC is that KIC is working correctly with Kong and is creating the right configuration in Kong or not. Mocking Kong's Admin API defeats that purpose.

PR #35, opened 2021-06-08
Add Github Actions Environment

shaneutt:
Codecov no longer allows us to anonymously report coverage for this repo, so we have to use Github Actions Environments going forward to include our codecov api key in the CI environment.


rainest (approved):
Don't see any reason not to approve, but any context around what we need the change for?

shaneutt:
> Don't see any reason not to approve, but any context around what we need the change for?

Sorry, I updated the description to explain.

Issue #10, opened 2021-03-23
Test Inception


shaneutt:
The testing framework needs to test itself.

Goals:
- [x] Test Coverage for KTF at least at 60%
- [x] Test Coverage Integration with Codecov added
- [x] Test Coverage button added to README.md

PR #42, opened 2021-07-11
chore: add codecov to ci

shaneutt:
Resolves https://github.com/Kong/kubernetes-testing-framework/issues/10


.github/workflows/test_integration_with_coverage.yaml
```diff
@@ -1,6 +1,14 @@
-name: Integration Tests
+name: Integration Tests & Coverage
 
-on: [pull_request]
+on:
+  pull_request:
+    branches:
+      - 'main'
```
mflendrich, line 6: optional

```suggestion
      - '*'
```

.github/workflows/test_unit.yaml
```diff
@@ -1,6 +1,14 @@
 name: Unit Tests
 
-on: [pull_request]
+on:
+  pull_request:
+    branches:
+      - 'main'
```
mflendrich, line 6: optional

```suggestion
      - '*'
```

Makefile
```diff
@@ -7,4 +7,5 @@ test.unit:
 
 .PHONY: test.integration
 test.integration:
-	@GOFLAGS="-tags=integration_tests" go test -race -v ./test/integration/...
+	@GOFLAGS="-tags=integration_tests" go test -race -v \
+		-covermode=atomic -coverprofile=coverage.out ./...
```
mflendrich, line 11: Is the change of the package path intentional? It doesn't seem so.

.github/workflows/test_integration_with_coverage.yaml
shaneutt: For the moment, I would like to only run this for main

Makefile
shaneutt: This is intentional. It makes sure that the integration tests also run the unit tests ahead of time so that we are reporting one single complete coverage.out.

rainest (approved):

PR #44, opened 2021-07-13
fix: use bash for install script

rainest:
Use bash for the install script instead of sh. The KTF install script
relies on some bash-specific shell builtins (set -o pipefail) and won't
run with plain sh.

This partially addresses the issue encountered in https://github.com/Kong/charts/runs/3060680578?check_suite_focus=true


ccfishk (approved):
LGTM

shaneutt (approved):

PR #45, opened 2021-07-14
chore: cleanup github actions workflows

shaneutt:
This patch combines all testing into a single workflow and
action, as well as coverage since this is all instrumented
with a single Makefile directive already anyway. Additionally
this patch flags the release workflow to wait for tests to
be passing on main before committing to a release of a tag.


.github/workflows/tests.yaml
```diff
@@ -27,10 +27,9 @@ jobs:
           ${{ runner.os }}-go
     - name: checkout repository
       uses: actions/checkout@v2
-    - name: run integration tests
+    - name: run unit and integration tests
       run: make test.integration
```
mflendrich, line 31:
```suggestion
      run: make test.unit test.integration
```

Given that you're removing a workflow that `make`s `test.unit`, don't we need to add `test.unit` here now that this step is supposed to run those?

mflendrich (approved):
LGTM blocked by resolution of the attached comment

.github/workflows/tests.yaml
shaneutt: We don't, the integration tests in this makefile run ALL tests, simply _including_ the integration tests (by way of go tag).

mflendrich: That's a misleading name for that target, then. Consider renaming to `test.all` alongside `test.integration` and `test.unit`

shaneutt: Sounds good, 09c27b175fd81ae78f7ebefa9bdbc36e5881b884 adds this :+1:

PR #58, opened 2021-07-23
[feat] add k8s version for environment builder

shaneutt:
Clusters already have the ability to designate a specific
Kubernetes version, this patch forwards that ability up
through the environment which creates the cluster and makes
the functionality available via CLI using the newly added
flag ktf env create --kubernetes-version <VERSION>.

In support of https://github.com/Kong/charts/issues/391

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@
 *.dll
 *.so
 *.dylib
-./ktf*
+ktf*
 
 # Test binary, built with `go test -c`
 *.test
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -4,7 +4,7 @@
 *.dll
 *.so
 *.dylib
-./ktf*
+ktf*
```
mflendrich, line 7: are you sure? what if you create a source file with a "ktf" prefix?

mflendrich (approved):

.gitignore
shaneutt: Yeah, I'm pretty sure we wont create files like that.

PR #81, opened 2021-08-04
fix: kong addon deployment idemopotence

shaneutt:
At times the repo add command might throw an error
even if ultimately the repo is already there.

This patch adds the `--force-overwrite` options when
adding the kong repo to update it to the latest value
regardless of whether or not it already existed.

Resolves https://github.com/Kong/kubernetes-testing-framework/issues/80

### DIFF
--- a/pkg/clusters/addons/kong/kong.go
+++ b/pkg/clusters/addons/kong/kong.go
@@ -112,7 +112,7 @@ func (a *Addon) Deploy(ctx context.Context, cluster clusters.Cluster) error {
 
 	// ensure the repo is added
 	stderr := new(bytes.Buffer)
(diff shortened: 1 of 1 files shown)


pkg/clusters/addons/kong/kong.go
```diff
@@ -112,7 +112,7 @@ func (a *Addon) Deploy(ctx context.Context, cluster clusters.Cluster) error {
 
 	// ensure the repo is added
```
mflendrich, line 113:
```suggestion
	// ensure the repo exists
```

small semantic difference - if we want the operation to be idempotent, we care about presence, not about addition.

mflendrich (approved):

PR #88, opened 2021-08-06
[wip] attach existing cluster for testing purpose

ccfishk:


shaneutt:
looks like this is still WIP, so I'm going to convert it to draft just so its clear :+1:

ccfishk:
existing EKS cluster through okta, saml2aws access for production only, could not attach. The expectation is not valid anymore.

PR #125, opened 2021-09-29
fix: re-enable integration tests parallelization

shaneutt:
Re-enables parallel integration tests: This was disabled previously because of problems found with Github Actions when spawning multiple clusters, but this patch includes a specific limit of 2 for that environment to try and reduce that problem but still squeeze out some extra time out of the tests there.

Additionally in this PR I fixed an issue with the Kong addon which I found to be causing test flakes for the enterprise integration tests: we were configuring the number of alphabetic characters required for a generated secret in place of the numbers of symbols required. This has been fixed and the result is passwords which are still long, but don't include some characters which require special escaping or handling (which would appear to be totally fine in the use case of testing). This fixed a test flake where the enterprise enabled Kong cluster would never come up in the tests due to failure to run the migrations and the test would time out.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -40,6 +40,8 @@ jobs:
       run: make test.integration
       env:
         KONG_ENTERPRISE_LICENSE: ${{ secrets.KONG_ENTERPRISE_LICENSE }}
+        NCPU: 2 # it was found that github actions (specifically) did not seem to perform well when spawning
+                # multiple kind clusters within a single job so this is hardcoded to 2 to ensure a limit of 2 clusters at any one point.
 
     - name: Upload coverage to Codecov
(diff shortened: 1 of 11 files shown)


.github/workflows/tests.yaml
```diff
@@ -40,6 +40,9 @@ jobs:
       run: make test.integration
       env:
         KONG_ENTERPRISE_LICENSE: ${{ secrets.KONG_ENTERPRISE_LICENSE }}
+        NCPU: 2 # it was found that github actions (specifically) did not seem to perform well when spawning
+                # multiple kind clusters within a single job so this is hardcoded to 2 to ensure a limit of 2 clusters at any one point.
+
```
shaneutt, line 45:
```suggestion
```

rainest (approved):

PR #149, opened 2021-11-12
Add support for getting node addresses and loading images

rainest:
Two separate features:

1. Add a new `GetNodeAddresses()` cluster function which returns a cluster's worker node addresses. Note that GKE does not allow access through these by default (at least not for NodePorts--unsure if they allow anything else by default).
2. Add a new loadimage plugin. It loads images into KIND clusters.

### DIFF
--- a/pkg/clusters/cluster.go
+++ b/pkg/clusters/cluster.go
@@ -46,4 +46,7 @@ type Cluster interface {
 
 	// DeleteAddon removes an existing cluster Addon.
 	DeleteAddon(ctx context.Context, addon Addon) error
+
+	// GetNodeAddresses returns a list of cluster worker node addresses
+	GetNodeAddresses(ctx context.Context) ([]string, error)
 }
--- a/pkg/clusters/types/gke/cluster.go
+++ b/pkg/clusters/types/gke/cluster.go
@@ -10,6 +10,8 @@ import (
 	container "cloud.google.com/go/container/apiv1"
 	"github.com/blang/semver/v4"
 	"google.golang.org/api/option"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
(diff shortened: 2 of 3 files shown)


pkg/clusters/addons/loadimage/addon.go
```diff
@@ -0,0 +1,101 @@
+package loadimage
```
shaneutt, line 1: Ultimately the goal of the cluster addons so far is to be usable on any `clusters.Cluster`, this addon is entirely tied to `kind` based clusters, breaking through the intended abstraction. It seems to me that we can move this same functionality as methods on the `kind` cluster type in `pkg/clusters/types/kind/` to isolate this to the cluster implementation that actually supports it, the effect of which is that you need to operate on it as it's underlying type (there's precedent for this in the addons). Let me know your thoughts?

pkg/clusters/cluster.go
```diff
@@ -46,4 +46,7 @@ type Cluster interface {
 
 	// DeleteAddon removes an existing cluster Addon.
 	DeleteAddon(ctx context.Context, addon Addon) error
+
+	// GetNodeAddresses returns a list of cluster worker node addresses
+	GetNodeAddresses(ctx context.Context) ([]string, error)
```
shaneutt, line 51: If this doesn't work on GKE then I don't feel we should be adding it to the cluster interface, rather we should add it to the kind cluster type and later we can promote it into the interface if we have a general cluster implementation.

shaneutt (changes requested):

pkg/clusters/addons/loadimage/addon.go
rainest: Which has the precedent?

pkg/clusters/cluster.go
rainest: It does work and is generally applicable to all clusters--everything does have worker nodes and those worker nodes have addresses--but GKE clusters require additional steps before the most obvious use case for this becomes viable: https://cloud.google.com/kubernetes-engine/docs/how-to/exposing-apps#creating_a_service_of_type_nodeport

I didn't try to implement that, however, as I had no need to do NodePort GKE testing, but ultimately if I did enabling access to the NodePorts would be independent of retrieving the node addresses.

pkg/clusters/addons/loadimage/addon.go
shaneutt: Sorry what I meant was that the addons do something similar to this, they each adhere to the interface but several of them have their own domain specific methods.

pkg/clusters/cluster.go
shaneutt: I gotcha. I think that we should add the configuration now if that seems feasible, if not let's at least make an issue. I'll consider this resolved as it doesn't need to block.

PR #150, opened 2021-11-15
feat: add Cluster node addresses getter method

rainest:
Split from https://github.com/Kong/kubernetes-testing-framework/pull/149

Add a new GetNodeAddresses() cluster function which returns a cluster's worker node addresses. Note that GKE does not allow access through these by default (at least not for NodePorts--unsure if they allow anything else by default). It does [still work as expected](https://github.com/Kong/kubernetes-testing-framework/pull/149#discussion_r748467326), I'm just not immediately sure how you'd use it. Node addresses are a property available on all clusters, however, regardless of their surrounding network and firewall rules, so this should go into the Cluster interface.

### DIFF
--- a/pkg/clusters/cluster.go
+++ b/pkg/clusters/cluster.go
@@ -46,4 +46,7 @@ type Cluster interface {
 
 	// DeleteAddon removes an existing cluster Addon.
 	DeleteAddon(ctx context.Context, addon Addon) error
+
+	// GetNodeAddresses returns a list of cluster worker node addresses
+	GetNodeAddresses(ctx context.Context) ([]string, error)
 }
--- a/pkg/clusters/types/gke/cluster.go
+++ b/pkg/clusters/types/gke/cluster.go
@@ -10,6 +10,8 @@ import (
 	container "cloud.google.com/go/container/apiv1"
 	"github.com/blang/semver/v4"
 	"google.golang.org/api/option"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
@@ -120,6 +122,22 @@ func (c *gkeCluster) Cleanup(ctx context.Context) error {
 
 func (c *gkeCluster) Client() *kubernetes.Clientset {
 	return c.client
+}
+
+func (c *gkeCluster) GetNodeAddresses(ctx context.Context) ([]string, error) {
+	var addrs []string
+	nodes, err := c.Client().CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return []string{}, err
+	}
+	for _, node := range nodes.Items {
+		for _, addr := range node.Status.Addresses {
+			if addr.Type == corev1.NodeExternalIP {
+				addrs = append(addrs, addr.Address)
+			}
+		}
+	}
+	return addrs, nil
 }
 
 func (c *gkeCluster) Config() *rest.Config {
--- a/pkg/clusters/types/kind/cluster.go
+++ b/pkg/clusters/types/kind/cluster.go
@@ -7,6 +7,8 @@ import (
 	"strings"
 	"sync"
 
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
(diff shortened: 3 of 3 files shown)


pkg/clusters/cluster.go
```diff
@@ -46,4 +46,7 @@ type Cluster interface {
 
 	// DeleteAddon removes an existing cluster Addon.
 	DeleteAddon(ctx context.Context, addon Addon) error
+
+	// GetNodeAddresses returns a list of cluster worker node addresses
+	GetNodeAddresses(ctx context.Context) ([]string, error)
```
shaneutt, line 51: In order to make this more generic and therefore simpler to implement on all `Cluster` implementations uniformly I would suggest returning `[]corev1.NodeAddress` type instead of `[]string`. Having some implementations providing IP addresses with wildly different routing characteristics could become problematic.

shaneutt (commented):
I want to avoid blocking, but I would like to get your thoughts on providing the underlying `[]NodeAddress` instead of `[]string` before we merge this?

pkg/clusters/cluster.go
rainest: Do we need to indeed provide the NodeAddress or could we instead allow you to select the type of addresses returned? Should clients actually need to make that decision, or can we make a a reasonable assumption about which type of address to return (what this is trying to do)?

The actual address inside a NodeAddress [is already just a string](https://pkg.go.dev/k8s.io/api/core/v1#NodeAddress) and the type is the only other field present. You always ultimately extract a string to get whatever you're routing to, though that may be either a hostname or a string IP address.

Ideally I'd like the user to not need to worry about knowing the details of which type they need to use, i.e. `GetNodeAddresses()` is a generic "this will return an address you can use to route to a node; it's up to the cluster to decide which address that is". This should just be either `NodeInternalIP` for local Docker-based clusters (both KIND and minikube only list internal addresses, but they are routeable from the host machine) or `NodeExternalIP` for everything else.

Nodes that require internet routing and only provide `NodeExternalDNS` and `NodeExternalIP` are theoretically possible (though I don't know why they'd do that), but I wouldn't expect that to be an issue, since from the client perspective you can usually plug in either a hostname or address into a URL or whatever, and your language runtime and/or OS will take care of resolution if needed.

shaneutt: > Do we need to indeed provide the NodeAddress or could we instead allow you to select the type of addresses returned? 
> Should clients actually need to make that decision, or can we make a a reasonable assumption about which type of address to return (what this is trying to do)?
> The actual address inside a NodeAddress [is already just a string](https://pkg.go.dev/k8s.io/api/core/v1#NodeAddress) and the type is the only other field present. You always ultimately extract a string to get whatever you're routing to, though that may be either a hostname or a string IP address.
> Ideally I'd like the user to not need to worry about knowing the details of which type they need to use, i.e. `GetNodeAddresses()` is a generic "this will return an address you can use to route to a node; it's up to the cluster to decide which address that is". This should just be either `NodeInternalIP` for local Docker-based clusters (both KIND and minikube only list internal addresses, but they are routeable from the host machine) or `NodeExternalIP` for everything else.
> Nodes that require internet routing and only provide `NodeExternalDNS` and `NodeExternalIP` are theoretically possible (though I don't know why they'd do that), but I wouldn't expect that to be an issue, since from the client perspective you can usually plug in either a hostname or address into a URL or whatever, and your language runtime and/or OS will take care of resolution if needed.

The previous "getter" methods on the `Cluster` interface all dealt with specific kinds of metadata about the cluster that would only be resolved a single way, making implementations straightforward. This getter method not only makes an assumption about the address type independently for each implementation, but also provides that data in isolation from it's parent data (e.g. the actual `Node`) which I can see potentially being important later on.

Ultimately I'm not in favor of this approach because:

1. I believe setting a precedent for interface which makes more underlying assumptions could start us in a direction of less predictability and more subtle issues in `Cluster` implementations in the long run.
2. I am having trouble seeing the value of adding this to the interface because it doesn't seem like something that would be commonly needed by calling tests, it seems like it would be perfectly reasonable for the tests which do need this to just list the nodes through the cluster client.
3. I think given recent discussions we've had there's some question as to what the `Cluster` interface really should be doing. I would personally like to avoid changing it until we've had some more time to think on that and provide a better definition or implementation.

Given all these things I would suggest we leave this out in favor of callers just listing the nodes through the clients for now. That said, I'm also not prepared to block on this as I understand the convenience it provides. I trust your judgement and only ask that you consider my thoughts but resolve this comment as you see fit.

shaneutt (approved):

pkg/clusters/cluster.go
rainest: Looked at briefly porting this over to the KIC test but it'd need `verifyIngress()` to make cluster type decisions. So far we have nothing of the sort (we have access to `cluster.Type()` but never use it) in E2E, and only use it elsewhere to select addons (in suite_test).

Intuitively, that seems like what we'd want: test innards don't care about the cluster type as much as possible (because it's less code to maintain), and tests overall just use the type for high-level decisions (load additional addons for this type, skip this test on this type, etc.). E2E is in a bit of an odd position, however, where it's de facto KIND-only anyway (all the tests build the entire cluster and all install metallb).

The use of `Cluster` could use clarification--I don't know if it should be used only for interacting with KTF (to perform setup/cleanup/etc.) or also as a general source of cluster metadata. Clearer delineation of the expected responsibilities of KTF versus the test plans that use it would also be helpful.

Here we can get away with making a quasi-type decision with fuzzy logic around the types of address available: all nodes _must_ have an internal address, and in absence of an external address, you may as well try it, so we don't need the test to have its own `Cluster.Type()` switch. That'd look like https://gist.github.com/rainest/820d995cb91d803b4d9946386aa8b715 versus https://github.com/Kong/kubernetes-ingress-controller/pull/2005/files#diff-d7850eda46771ac1eb42d866bb557fd98bdea88b6bcfba6b26b10cc975a40856R338-R357

rainest:
Handling in the test by using the client directly instead: https://github.com/Kong/kubernetes-ingress-controller/blob/caafc4e41ecd2c400ebcf0ed4175dd1132daf8ff/test/e2e/all_in_one_test.go#L374-L394

Issue #147, opened 2021-11-09
A registry addon


hutchic:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem Statement

As someone developing Kong
I want the ability to run a bespoke locally built docker image
So that I can test my code locally and within CI

### Proposed Solution

Prior art of running a local docker registry and connecting it into kind exists [here](https://github.com/Kong/khcp/blob/master/charts/local.sh#L1-L44)

Generally speaking taking a flag `--gateway-image=myimage` that gets from my local docker images and is the image KTF uses to run the gateway

shaneutt:
This has additional benefit for CI because it could allow for seeding images from a cache in Github actions, we should see if we can _also_ build this for a use case where the caller would have simple calls to seed the registry with multiple images.

PR #151, opened 2021-11-15
feat: add load image addon

rainest:
This is one of two proposals to add image loading to KTF, broken out of #149. It is mutually exclusive with https://github.com/Kong/kubernetes-testing-framework/pull/152. Only one of the two should be merged.

This approach provides a plugin for loading images following the standard KTF plugin pattern. Building a cluster that includes this plugin will attempt to load an image and report a failure if that is not possible or if that is unsuccessful.

This addition lacks tests because KIND will not (contrary to my earlier expectations) load an image not already present on the host machine. We'd need to run `docker pull` or similar to ensure the requested image is available locally, and I'm loathe to screw around with that. My original test code is available at https://gist.github.com/rainest/079d07206913398acf65371703cecf0d

**Pro/con evaluation**

The argument against this approach is that we should not add cluster addons unless they are universally available on all clusters types.

I am in favor of this option:

- I don't see an obvious reason to impose this restriction so long as we can provide feedback indicating that an addon is unavailable. While this case deals with cluster types, it seems reasonable that we'd extend the same paradigm to cluster versions (e.g. a GWAPI addon would not be available on 1.16, but would be available on {future version}). Addons are by definition not mandatory, and users can always simply not include them if they are unavailable on the desired cluster type.
- Using the addon pattern allows us to establish a standard means of requesting an addon reporting availability that works regardless of your target cluster. We currently [return an error in `Deploy()`](https://github.com/Kong/kubernetes-testing-framework/blob/67eb46f5f7dbf0794287f90edfec18beb8062d31/pkg/clusters/addons/metallb/metallb.go#L53) for unavailability. Code that utilizes these is cluster agnostic: you always call `WithAddons(<instance>)`, which always reports any failures when calling `Build()`. If you have multiple cluster types that support an addon set, swapping out the cluster type only requires calling a different package's `New()` function and passing the resulting `Cluster` downstream. If you go with #152's approach, any test that requires a non-universal plugin must be written for the specific target cluster type.
- In this specific case, loading images is not strictly specific to KIND. Although we don't yet provide a Minikube Cluster implementation, it supports [an almost identical command](https://minikube.sigs.k8s.io/docs/handbook/pushing/#7-loading-directly-to-in-cluster-container-runtime). Loading images into GKE arguably isn't impossible, it's just much more complicated (you'd need to ensure GCR is enabled, push an image to it, and provide the new GCR image name to the user) and not something I needed for now.
- Although I can't think of an example offhand, it seems reasonable that we may create addons that have partial support for some clusters, and the `Build() -> Deploy() -> succeed or fail based on availability` paradigm seems like it'd handle this easily.

We should look at more robust means of handling support indications if we think `Deploy()` failures are insufficient, but I think having a generic interface for attempting to load addons is far preferable to having separate paradigms for similar tasks (enabling functionality not exposed by the base Cluster functions), one of which is incapable of sharing both implementation and user-written test code.

### DIFF
--- a/pkg/clusters/addons/loadimage/addon.go
+++ b/pkg/clusters/addons/loadimage/addon.go
@@ -37,11 +37,6 @@ func (a *Addon) Name() clusters.AddonName {
 }
 
 func (a *Addon) Deploy(ctx context.Context, cluster clusters.Cluster) error {
-	if a.image == "" {
-		a.loaded = true
-		return nil
-	}
-
 	switch ctype := cluster.Type(); ctype {
 	case kind.KindClusterType:
 		if err := a.loadIntoKind(cluster); err != nil {
--- a/pkg/clusters/addons/loadimage/addon.go
+++ b/pkg/clusters/addons/loadimage/addon.go
@@ -55,8 +55,16 @@ func (a *Addon) Deploy(ctx context.Context, cluster clusters.Cluster) error {
 }
 
 func (a *Addon) Delete(ctx context.Context, cluster clusters.Cluster) error {
-	// per https://github.com/kubernetes-sigs/kind/issues/658 this is basically impossible
-	return fmt.Errorf("cannot remove loaded images from a cluster")
+	switch ctype := cluster.Type(); ctype {
+	case kind.KindClusterType:
+		// per https://github.com/kubernetes-sigs/kind/issues/658 this is basically impossible
+		// we lie here, because we want to mask this error. not deleting an image from KIND is benign:
+		// you either don't use it after (in which case you shouldn't care that it's still present) or
+		// load another image with the same name (in which case the name will point to the new image)
+		return nil
+	default:
+		return fmt.Errorf("loadimage addon is not supported by cluster type '%v'", cluster.Type())
+	}
 }
 
 func (a *Addon) Ready(ctx context.Context, cluster clusters.Cluster) ([]runtime.Object, bool, error) {
--- a/pkg/clusters/addons/loadimage/addon.go
+++ b/pkg/clusters/addons/loadimage/addon.go
@@ -1,0 +1,101 @@
+package loadimage
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"os/exec"
+
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/kong/kubernetes-testing-framework/pkg/clusters"
+	"github.com/kong/kubernetes-testing-framework/pkg/clusters/types/kind"
+)
+
+// -----------------------------------------------------------------------------
+// CertManager Addon - Builder
+// -----------------------------------------------------------------------------
+
+type Builder struct {
+	image string
+}
+
+func NewBuilder() *Builder {
+	return &Builder{}
+}
+
+func (b *Builder) WithImage(image string) *Builder {
+	b.image = image
+	return b
+}
+
+func (b *Builder) Build() *Addon {
+	return &Addon{
+		image:  b.image,
+		loaded: false,
+	}
+}
+
+// -----------------------------------------------------------------------------
+// CertManager Addon
+// -----------------------------------------------------------------------------
+
+const (
+	// AddonName indicates the unique name of this addon.
+	AddonName clusters.AddonName = "loadimage"
+)
+
+type Addon struct {
+	image  string
+	loaded bool
+}
+
+func New() clusters.Addon {
+	return &Addon{}
+}
+
+// -----------------------------------------------------------------------------
+// CertManager Addon - Addon Implementation
+// -----------------------------------------------------------------------------
+
+func (a *Addon) Name() clusters.AddonName {
+	return AddonName
+}
+
+func (a *Addon) Deploy(ctx context.Context, cluster clusters.Cluster) error {
+	if a.image == "" {
+		a.loaded = true
+		return nil
+	}
+	if cluster.Type() != kind.KindClusterType {
+		return fmt.Errorf("addon %v is not available on cluster types other than %v", a.Name(), kind.KindClusterType)
+	}
+	deployArgs := []string{
+		"load", "docker-image",
+		a.image,
+		"--name", cluster.Name(),
+	}
+
+	stderr := new(bytes.Buffer)
+	cmd := exec.Command("kind", deployArgs...)
+	cmd.Stdout = io.Discard
+	cmd.Stderr = stderr
+
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("%s: %w", stderr.String(), err)
+	}
+	a.loaded = true
+
+	return nil
+}
+
+func (a *Addon) Delete(ctx context.Context, cluster clusters.Cluster) error {
+	// per https://github.com/kubernetes-sigs/kind/issues/658 this is basically impossible
+	return fmt.Errorf("cannot remove loaded images from a cluster")
+}
+
+func (a *Addon) Ready(ctx context.Context, cluster clusters.Cluster) ([]runtime.Object, bool, error) {
+	// no way to verify this, we rely on Deploy's cmd.Run() not failing
+	return nil, a.loaded, nil
+}
--- a/pkg/clusters/addons/loadimage/addon.go
+++ b/pkg/clusters/addons/loadimage/addon.go
@@ -1,41 +1,14 @@
 package loadimage
 
 import (
-	"bytes"
 	"context"
 	"fmt"
-	"io"
-	"os/exec"
 
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/kong/kubernetes-testing-framework/pkg/clusters"
 	"github.com/kong/kubernetes-testing-framework/pkg/clusters/types/kind"
 )
-
-// -----------------------------------------------------------------------------
-// CertManager Addon - Builder
-// -----------------------------------------------------------------------------
-
-type Builder struct {
-	image string
-}
-
-func NewBuilder() *Builder {
-	return &Builder{}
-}
-
-func (b *Builder) WithImage(image string) *Builder {
-	b.image = image
-	return b
-}
-
-func (b *Builder) Build() *Addon {
-	return &Addon{
-		image:  b.image,
-		loaded: false,
-	}
-}
 
 // -----------------------------------------------------------------------------
 // CertManager Addon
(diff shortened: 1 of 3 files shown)


pkg/clusters/addons/loadimage/addon.go
```diff
@@ -0,0 +1,101 @@
...
+	b.image = image
+	return b
+}
+
+func (b *Builder) Build() *Addon {
+	return &Addon{
+		image:  b.image,
+		loaded: false,
+	}
+}
```
mflendrich, line 38: for consistency with other addons, the builder implementation should probably live in `builder.go`

pkg/clusters/addons/loadimage/addon.go
```diff
@@ -0,0 +1,101 @@
...
+func (a *Addon) Name() clusters.AddonName {
+	return AddonName
+}
+
+func (a *Addon) Deploy(ctx context.Context, cluster clusters.Cluster) error {
+	if a.image == "" {
+		a.loaded = true
+		return nil
+	}
+	if cluster.Type() != kind.KindClusterType {
```
mflendrich, line 71: consider using the Strategy pattern - having several implementations (for KIND, future minikube, etc) implementing a common interface. The concrete implementation to be used could be decided upon by the builder or by `Deploy()` whatever makes more sense.

pkg/clusters/addons/loadimage/addon.go
rainest: 361f92f

rainest: 361f92f

pkg/clusters/addons/loadimage/addon.go
rainest: 361f92f though with just a switch. I don't see anything obvious we'd add to an interface other than a single "load this" function, whose implementation is opaque to the rest of the addon.

shaneutt: If I'm understanding you correctly that you're suggesting we implement the strategy pattern for `Clusters` and to allow some cluster implementations to decide whether they have the capability to use the LoadImage addon, as the inverse to letting the addon make that determination, I do generally like the idea.

However, I'm very hesitant to put the load of doing that on this one addon, especially given that this addon follows an existing precedent, because I think this is something we need to look at holistically and outside of the immediate needs which drove this implementation.

I am OK with us using this simple logic for now, if we create and issue which describes the need @rainest and I have been talking about in zooms recently for clusters to have a system to describe optional cabilities and suggest the strategy pattern there, followed by linking that issue in a comment here and on metallb to point out that there's more TODO here.

shaneutt: If I'm understanding you correctly that you're suggesting we implement the strategy pattern for `Clusters` and to allow some cluster implementations to decide whether they have the capability to use the LoadImage addon, as the inverse to letting the addon make that determination, I do generally like the idea.

However, I'm very hesitant to put the load of doing that on this one PR, especially given that this addon follows an existing precedent, because I think this is something we need to look at holistically and outside of the immediate needs which drove this implementation.

I am OK with us using this simple logic for now, if we create and issue which describes the need @rainest and I have been talking about in zooms recently for clusters to have a system to describe optional cabilities and suggest the strategy pattern there, followed by linking that issue in a comment here and on metallb to point out that there's more TODO here.

shaneutt:
thinking more about this over the morning it seems like having a cluster-deployed registry option could solve this problem reasonably well, and has the following benefits over this implementation:

- would work on _every_ type of cluster with a single implementation
- makes the source of the image must less dependent on a local docker installation
- is already an active feature request in #149 

Let me know your thoughts?

rainest:
Re https://github.com/Kong/kubernetes-testing-framework/pull/151#issuecomment-971699443 tl;dr nah, use this still and add a push to GCR option if you want the same on GKE later. The experience for the test client is the same, and it's simpler and more standard.

Accessing images currently in the local Docker instance is the point here: if the Docker image was already published, you wouldn't need to load the image at all; you'd just provide the public URL of the image. This isn't a problem for the intended use case (you don't use this unless you have an image you haven't published), it's only a problem for running tests: in CI, you need to build the image before you can push it. Using an alternative private registry doesn't remove the requirement that you push images to the registry in advance.

Building and pushing is the responsibility of the test client, since you're (presumably) using this to test against an unpublished version of the application you're using KTF to test (KIC). I didn't set that up for https://github.com/Kong/kubernetes-ingress-controller/pull/2005 since running the test locally against the development version seemed sufficient before those tests would normally run (after a tag push builds an image and a version with these changes _is_ publicly available). I think having KIC CI build the image and run E2E against it would be useful, but it's something we don't strictly need immediately and the development time to add that also given that the goal here is to work around issues using the integration tests (no established paradigm of running differently-configured clusters during the same test run) or unit tests (nigh impossible here) for this.

The system mentioned in #147 is [KIND's local registry option](https://kind.sigs.k8s.io/docs/user/local-registry/). That relies on the registry and cluster running on the same host, with Docker network assistance to expose one to the other. It wouldn't work in other contexts without a lot of hackery. Cluster-deployed registries that the cluster then pulls from aren't impossible, but there doesn't appear to be a well-established practice around doing it, just several blog articles that indicate it's doable. Given the amount of setup involved though, I'm not convinced that it's valuable--the end goal is the same, and for hosted clusters it's almost certainly going to be easier to just push to the standard hosted registry for that provider. This maybe goes back to the philosophical dispute we have over how this should be used, but IMO heterogeneous environments are a fact of life in IT and we should contend with that reality; you don't get everything working the same way universally unless you're in physics instead.

pkg/clusters/addons/loadimage/addon.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (a *Addon) Delete(ctx context.Context, cluster clusters.Cluster) error {
+	// per https://github.com/kubernetes-sigs/kind/issues/658 this is basically impossible
+	return fmt.Errorf("cannot remove loaded images from a cluster")
+}
+
+func (a *Addon) Ready(ctx context.Context, cluster clusters.Cluster) ([]runtime.Object, bool, error) {
+	// no way to verify this, we rely on Deploy's cmd.Run() not failing
+	return nil, a.loaded, nil
+}
```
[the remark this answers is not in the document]
rainest: This is the first addon where `Deploy()` and `Ready()` are atomic: the others are all Deployments are similar where you get confirmation that the request to create succeeded, but readiness depends on an async process after that you must confirm separately.

Here, once `kind load docker-image` returns successfully, we're assured that the image is available (barring a hypothetical very odd bug in the command itself). That pattern should remain the same in any other implementations I can think of (absent a non-registry push like this, you're pushing to some registry, and `docker push` only succeeds if it actually pushed the image). My comment isn't correct--we actually can verify readiness, we just know it from earlier rather than needing to perform some additional check.

As a general pattern for atomic deploys, do you think this approach is sound? I think it's probably superfluous (AFAIK the `Deploy()` error will bubble up and test clients should abort then unless they ignore), but it otherwise still maintains the contract: `Ready()` does still report whether the addon is ready.

Interrogating containerd wouldn't be possible on registry-based solutions unless there's some way to list images on a remote registry (you won't be able to ssh into the kubelets to run crictl). That isn't a standard Docker feature (though a request has been open since 2015!), though it may be possible with individual registry providers through vendor-specific APIs, e.g. GCR [does provide this](https://cloud.google.com/container-registry/docs/managing#listing_images_by_their_storage_location).

pkg/clusters/addons/loadimage/addon.go
```diff
@@ -0,0 +1,65 @@
...
+	default:
+		return fmt.Errorf("loadimage addon is not supported by cluster type '%v'", cluster.Type())
+	}
+
+	return nil
+}
+
+func (a *Addon) Delete(ctx context.Context, cluster clusters.Cluster) error {
+	// per https://github.com/kubernetes-sigs/kind/issues/658 this is basically impossible
+	return fmt.Errorf("cannot remove loaded images from a cluster")
```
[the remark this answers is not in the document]
rainest: "what addons are today" very much needs a definition to be a useful bar if it's something beyond satisfying the interface, which appears to be the implication here 🙂 My intuitive definition was "almost anything we add to the test cluster after it starts"; the implied definition appears to instead be "things you can add by applying a Kubernetes manifest, nothing else", or maybe a slightly broader "things that function the same way on all Kubernetes implementations". If it's one of the latter two, IMO KTF needs some new, similar interface for things that aren't manifests, but also aren't something we'd add to the Cluster interface instead, without passing the buck for dealing with all implementation idiosyncrasies downstream to test clients.

We can't shield them from _all_ idiosyncrasies, but we can probably reasonably shield them from most. In this specific case, you shouldn't much care about deleting Docker images; you can just upload a new one with the same name to replace it. Some domain-specific knowledge of what the addon is doing will probably be necessary unless you really constrain what addons are.

I considered this fine because I wasn't aware of any scenarios where you'd call an addon's `Delete()` without knowing the specific addon, and if you did try to delete this one during the course of development, you'd learn it's not possible on the first attempt and not try it afterwards (or be so determined to purge the image that you contribute `kind delete docker-image` upstream).

Currently we don't call the `Delete()`s anywhere, so there's no practical use cases to draw from. This would fail if you did a blind `ListAddons()` and then `DeleteAddon()` on all results that aborted on any failure. That's only viable if we say "don't do that" (honestly, if you want that delete the cluster and make a new one; eat the extra time) or create the suggested mechanism that reports a sort of benign non-error failure.

It would be possible in other implementations, e.g. [GCR does let you delete images](https://cloud.google.com/container-registry/docs/managing#deleting_images), but that doesn't remove the issue that some cluster types will always fail.

25f0846 splits the cases and makes KIND's delete a noop.

shaneutt (approved):
approving based on a desire to keep things unblocked, feel free to resolve any open comments as you see fit. I can accept the current implementation for now, despite wanting some different things.

pkg/clusters/addons/loadimage/addon.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (a *Addon) Name() clusters.AddonName {
+	return AddonName
+}
+
+func (a *Addon) Deploy(ctx context.Context, cluster clusters.Cluster) error {
+	if a.image == "" {
+		a.loaded = true
+		return nil
+	}
```
mflendrich, line 43: Why is this here - why not consider an unset image a validation error instead? The latter feels like a better failure mode to me.
If there's some explanation that I've missed, I suggest that we explain the decision in a comment.

mflendrich (approved):

pkg/clusters/addons/loadimage/addon.go
rainest: 01f4ed0 and daa1374

The idea was sorta that you can successfully load no image by doing nothing. This allowed for some laziness downstream, which originally called this with no conditions:

```
	extraImagePlugin := loadimage.NewBuilder().WithImage(os.Getenv(imageOverrideEnvVar)).Build()
	builder := environments.NewBuilder().WithAddons(extraImagePlugin)
```
That envvar may be unset, and the noop version allowed you to not check anything and run with it anyway. In retrospect, more confusing to read that downstream anyway, so now it actually errors and I'll make downstream check what it should do more strictly.

PR #152, opened 2021-11-15
feat: add load image KIND helper

rainest:
This is one of two proposals to add image loading to KTF, broken out of #149. It is mutually exclusive with https://github.com/Kong/kubernetes-testing-framework/pull/151. Only one of the two should be merged. Pro/con discussion is in #151.

This approach makes `LoadImage()` a function in the KTF `kind` package. Ensuring that the test cluster is a KIND cluster and that the named cluster is the KIND cluster used for testing is delegated to the library user.

This addition lacks tests because KIND will not (contrary to my earlier expectations) load an image not already present on the host machine. We'd need to run `docker pull` or similar to ensure the requested image is available locally, and I'm loathe to screw around with that.

### DIFF
--- a/pkg/clusters/types/kind/cluster.go
+++ b/pkg/clusters/types/kind/cluster.go
@@ -1,9 +1,12 @@
 package kind
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"os"
+	"os/exec"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


mflendrich:
since you've asked for feedback: I'm "thumbs up" for #151 and "neutral" for this. Added some comments in #151.

shaneutt:
> since you've asked for feedback: I'm "thumbs up" for #151 and "neutral" for this. Added some comments in #151.

@mflendrich can you help me to better understand your reasoning for #151 vs this option?

rainest:
Per sync discussion, we'll proceed with #151

PR #192, opened 2022-01-05
Bump Enterprise to 2.7 and release 0.11.1

rainest:
Needed to fix https://github.com/Kong/kubernetes-ingress-controller/issues/2126

Tests that delete and restore their upstreams in non-default workspaces may be affected by the upstream event fix in https://docs.konghq.com/gateway/changelog/#core-6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.11.1
+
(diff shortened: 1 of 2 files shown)


shaneutt (approved):
Cool, hopefully this takes care of it :crossed_fingers:

PR #196, opened 2022-01-19
feat(knative) support user-supplied versions

rainest:
Support user-supplied Knative versions, and default to the latest release at this time.

Basic PoC using the controller integration tests:

[knative_test.txt](https://github.com/Kong/kubernetes-testing-framework/files/7899557/knative_test.txt)

```
10:08:55-0800 esenin $ kubectl describe po -n knative-serving | grep release=
                serving.knative.dev/release=v1.1.1
```

### DIFF
--- a/pkg/clusters/addons/knative/knative.go
+++ b/pkg/clusters/addons/knative/knative.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/kong/kubernetes-testing-framework/pkg/clusters"
+	"github.com/kong/kubernetes-testing-framework/pkg/utils/github"
 )
 
 // -----------------------------------------------------------------------------
@@ -29,7 +30,7 @@ const (
 	DefaultNamespace = "knative-serving"
 
 	// DefaultVersion is the Knative version deployed when the user requests no specific version
-	DefaultVersion = "knative-v1.1.1"
+	DefaultVersion = "0.0.0"
 )
 
 type addon struct {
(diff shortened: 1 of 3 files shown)


pkg/clusters/addons/knative/knative.go
```diff
@@ -27,12 +27,17 @@ const (
 
 	// DefaultNamespace indicates the default namespace this addon will be deployed to.
 	DefaultNamespace = "knative-serving"
+
+	// DefaultVersion is the Knative version deployed when the user requests no specific version
+	DefaultVersion = "knative-v1.1.1"
```
shaneutt, line 32: we have a mixture of addons either having a default version or deriving a base version from github releases, e.t.c.

Now that knative is at V1, it seems like perhaps we should just use the existing utility function to base our default/base version off of the latest release?

shaneutt (commented):
CI is having a failure:

```
knative core cleanup failed STDOUT=() STDERR=(error: the URL passed to filename "https://github.com/knative/serving/releases/download/%s/serving-core.yaml" is not valid: parse "https://github.com/knative/serving/releases/download/%s/serving-core.yaml": invalid URL escape "%s/"
```

pkg/clusters/addons/knative/knative.go
rainest: 92056b9

pkg/clusters/addons/knative/knative.go
```diff
@@ -221,3 +231,14 @@ func deleteKnative(ctx context.Context, cluster clusters.Cluster) error {
 		}
 	}
 }
+
+// useLatestKnativeVersion locates and sets the istio version to deploy to the latest
+// non-prelease tag found.
```
shaneutt, line 236:
```suggestion
// useLatestKnativeVersion locates and sets the istio version to deploy to the latest
// non-prelease tag found.
```
```suggestion
// useLatestKnativeVersion locates and sets the knative version to deploy to the latest
// non-prelease tag found.
```

shaneutt (approved):
LGTM after one more suggestion I've added is resolved and the CI failures are fixed.

PR #209, opened 2022-01-28
fix(knative) add retry to core deployment

### LINKED ISSUE #208: Knative install may fail due to API slowness

### Current Behavior

Knative is installed in two components, CRDs and Core:

https://github.com/Kong/kubernetes-testing-framework/blob/269c1198aa5ddb5f1fbf395a2a1e87a122f0b695/pkg/clusters/addons/knative/knative.go#L107-L123

Note that the `--wait` here isn't actually helping, since it only applies to deletions with finalizers.

The Core contains CRs using the CRDs. Failing test logs indicate that although we install the CRDs first, and the API server even acknowledges their presence (the unchanged here is from the Core install, whose manifest includes the CRDs also--the CRDs install succeeded and didn't log its output), that they are not yet truly available:

```
2022-01-28T15:36:47.1793234Z customresourcedefinition.apiextensions.k8s.io/images.caching.internal.knative.dev unchanged
2022-01-28T15:36:47.1820005Z ) STDERR=(error: unable to recognize "https://github.com/knative/serving/releases/download/knative-v1.2.0/serving-core.yaml": no matches for kind "Image" in version "caching.internal.knative.dev/v1alpha1"
```

### Expected Behavior

The install succeeds.

We should be able to brute-force it by re-running the core apply if it fails with a timeout, a la the `require.Eventually()` in tests. However, this is outside tests, and doesn't have access to that exact function; we need to build our own.

### Steps To Reproduce

```markdown
Random based on CI timing.
```

### Kong Kubernetes Testing Framework Version

```shell
Anything that includes https://github.com/Kong/kubernetes-testing-framework/pull/196. Older Knative versions were apparently lucky enough not to trigger the race.
```

### Kubernetes version

_No response_

### Anything else?

_No response_

rainest:
Fix #208 using something sort of like `require.Eventually()`

Fix is speculative based on the observed issue. We have no guaranteed means of reproducing the issue to confirm.

### DIFF
--- a/pkg/clusters/addons/knative/knative.go
+++ b/pkg/clusters/addons/knative/knative.go
@@ -94,6 +94,8 @@ func (a *addon) Ready(ctx context.Context, cluster clusters.Cluster) ([]runtime.
 const (
 	knativeCRDs = "https://github.com/knative/serving/releases/download/%s/serving-crds.yaml"
 	knativeCore = "https://github.com/knative/serving/releases/download/%s/serving-core.yaml"
+	retryCount  = 5
+	retryWait   = 10
 )
 
 func deployKnative(ctx context.Context, cluster clusters.Cluster, version string) error {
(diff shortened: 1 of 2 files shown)


pkg/clusters/addons/knative/knative.go
```diff
@@ -114,12 +116,20 @@ func deployKnative(ctx context.Context, cluster clusters.Cluster, version string
...
 
 	// apply the core deployments, but don't wait because we're going to patch them
-	cmd = exec.CommandContext(ctx, "kubectl", "--kubeconfig", kubeconfig.Name(), "apply", "-f", fmt.Sprintf(knativeCore, version)) //nolint:gosec
-	stdout, stderr = new(bytes.Buffer), new(bytes.Buffer)
-	cmd.Stdout = stdout
-	cmd.Stderr = stderr
-	if err := cmd.Run(); err != nil {
-		return fmt.Errorf("knative core deployment failed STDOUT=(%s) STDERR=(%s): %w", stdout.String(), stderr.String(), err)
+	// the CRDs applied earlier may not become available immediately, so retry this several times if it fails
+	for i := range make([]int, retryCount) {
```
shaneutt, line 120:
```suggestion
	for i := 0 ; i < retrycount ; i++ {
```

shaneutt (approved):

PR #222, opened 2022-02-17
feat(kind) let users specify a KIND config file

rainest:
As seen in https://github.com/Kong/kubernetes-ingress-controller/pull/2258

When creating a KIND cluster, users can specify a KIND configuration file. This requires creating a cluster and then an environment with that existing cluster, since it is not a generic Environment feature.


pkg/clusters/types/kind/builder.go
```diff
@@ -42,13 +43,23 @@ func (b *Builder) WithClusterVersion(version semver.Version) *Builder {
 	return b
 }
 
+// WithConfig sets a filename containing a KIND config
+func (b *Builder) WithConfig(filename string) *Builder {
```
shaneutt, line 47: Seems like a good place to drop a helpful link for future code readers:

```suggestion
// WithConfig sets a filename containing a KIND config
// See: https://kind.sigs.k8s.io/docs/user/configuration/
func (b *Builder) WithConfig(filename string) *Builder {
```

shaneutt (approved):

shaneutt:
We do need to make sure that we test this locally before merging, then LGTM.

shaneutt:
Sounds good :vulcan_salute: :ship:

PR #253, opened 2022-04-11
feat: expose default TCP port for Kong addon

shaneutt:
Exposes the default TCP routing port for the Kong Addon as a far, similar to what already exists for UDP. Also adds references to the var instead of literals where applicable.

Needed for https://github.com/Kong/kubernetes-ingress-controller/issues/2086

### DIFF
--- a/pkg/clusters/addons/kong/vars.go
+++ b/pkg/clusters/addons/kong/vars.go
@@ -45,7 +45,7 @@ const (
 
 	// DefaultTCPServicePort indicates the default open port that will be used for TCP traffic.
 	DefaultTCPServicePort = 8888
-	
+
 	// DefaultTLSServicePort indicates the default open port that will be used for TLS traffic.
 	DefaultTLSServicePort = 8899
 )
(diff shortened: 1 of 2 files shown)


pkg/clusters/addons/kong/vars.go
```diff
@@ -42,4 +42,7 @@ const (
 
 	// DefaultUDPServicePort indicates the default open port to be found on the Kong proxy's UDP service.
 	DefaultUDPServicePort = 9999
+
+	// DefaultUDPServicePort indicates the default open port that will be used for TCP traffic.
+	DefaultTCPServicePort = 8888
```
rainest, line 47: While we're here, can we add DefaultTLSServicePort 8899 in `kong/addon.go` also, assuming we'll need it imminently for TLSRoute?

Once again, I am mad at GH for not allowing suggestions to extend onward into code outside the diff region.

```suggestion
	// DefaultTCPServicePort indicates the default open port that will be used for TCP traffic.
	DefaultTCPServicePort = 8888
	
	// DefaultTLSServicePort indicates the default open port that will be used for TLS traffic.
	DefaultTLSServicePort = 8899
```

rainest (changes requested):

rainest:
Shoved https://github.com/Kong/kubernetes-testing-framework/pull/253/commits/84bb5b82d6edf2f5842cfb2bfb45a438a75a49b6 in as the second part of the suggestion that I couldn't actually suggest. YOLO approving my own commits revert if needed before release.

rainest (approved):

PR #273, opened 2022-05-23
fix: remove cluster init token check

shaneutt:
Back when we first made KTF we added a check to wait for the service account that would be used for testing until it's `Secret` was generated. Modern versions of Kubernetes since `v1.24.0` have a different implementation for `ServiceAccounts` which doesn't make this secret available like it used to be. As such this patch removes the check as it is irrelevant going forward and is breaking recent deployments.

### DIFF
--- a/internal/utils/cluster_init.go
+++ b/internal/utils/cluster_init.go
@@ -55,25 +55,6 @@ func ClusterInitHooks(ctx context.Context, cluster clusters.Cluster) error {
 		}
 	}
 
-	// ensure the secret token for the default sa is available
-	var defaultSATokenReady bool
-	for !defaultSATokenReady {
-		select {
-		case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


shaneutt:
> So the wait is not necessary anymore?

The wait was probably never really necessary in the first place, but it's definitely not necessary now. In fact, it simply _won't_ work because the `Secret` is no longer made.

mflendrich (approved):

mflendrich:
Did some research:

[This KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/1205-bound-service-account-tokens#acls-for-tokenrequest) says

> The NodeAuthorizer will allow the kubelet to use its credentials to request a service account token on behalf of pods running on that node.

Which means that getting SA credentials for a pod is now a synchronous operation performed by kubelet.

PR #288, opened 2022-06-08
feat: add kong-ingress-controller-image flag

jrsmroz:
This PR adds kong-ingress-controller-image that sets the ingress controller image used for deployment.

### DIFF
--- a/internal/cmd/ktf/environments.go
+++ b/internal/cmd/ktf/environments.go
@@ -50,6 +50,7 @@ func init() { //nolint:gochecknoinits
 	environmentsCreateCmd.PersistentFlags().StringArray("addon", nil, "name of an addon to deploy to the testing environment's cluster")
 	environmentsCreateCmd.PersistentFlags().Bool("kong-disable-controller", false, "indicate whether the kong addon should have the controller disabled (proxy only)")
 	environmentsCreateCmd.PersistentFlags().Bool("kong-admin-service-loadbalancer", false, "indicate whether the kong addon should deploy the proxy admin service as a LoadBalancer type")
+	environmentsCreateCmd.PersistentFlags().String("kong-ingress-controller-image", "", "use a specific ingress controller container image for the Gateway (proxy)")
 	environmentsCreateCmd.PersistentFlags().String("kong-gateway-image", "", "use a specific container image for the Gateway (proxy)")
 	environmentsCreateCmd.PersistentFlags().String("kong-dbmode", "off", "indicate the backend dbmode to use for kong (default: \"off\" (DBLESS mode))")
 }
@@ -210,12 +211,26 @@ func configureKongAddon(cmd *cobra.Command, envBuilder *environments.Builder) *e
 	if customGatewayImage != "" {
 		imageParts := strings.Split(customGatewayImage, ":")
 		if len(imageParts) == 1 {
-			imageParts[1] = "latest"
+			imageParts = append(imageParts, "latest")
 		}
 		if len(imageParts) != 2 { //nolint:gomnd
 			cobra.CheckErr(fmt.Errorf("malformed --kong-gateway-image: %s", customGatewayImage))
 		}
 		builder.WithProxyImage(imageParts[0], imageParts[1])
+	}
+
+	customControllerImage, err := cmd.PersistentFlags().GetString("kong-ingress-controller-image")
+	cobra.CheckErr(err)
+
+	if customControllerImage != "" {
+		imageParts := strings.Split(customControllerImage, ":")
+		if len(imageParts) == 1 {
+			imageParts = append(imageParts, "latest")
+		}
+		if len(imageParts) != 2 { //nolint:gomnd
+			cobra.CheckErr(fmt.Errorf("malformed --kong-ingress-controller-image: %s", customControllerImage))
+		}
+		builder.WithControllerImage(imageParts[0], imageParts[1])
 	}
 
 	enableAdminSvcLB, err := cmd.PersistentFlags().GetBool("kong-admin-service-loadbalancer")
--- a/pkg/clusters/addons/kong/builder.go
+++ b/pkg/clusters/addons/kong/builder.go
@@ -117,8 +117,7 @@ func (b *Builder) WithProxyImage(repo, tag string) *Builder {
 	return b
 }
 
-// WithControllerImage configures the ingress controller
-// container image name and tag for the Kong proxy.
+// WithControllerImage configures the ingress controller container image name and tag.
 func (b *Builder) WithControllerImage(repo, tag string) *Builder {
 	b.ingressControllerImage = repo
 	b.ingressControllerImageTag = tag
--- a/pkg/clusters/addons/kong/builder.go
+++ b/pkg/clusters/addons/kong/builder.go
@@ -21,6 +21,8 @@ type Builder struct {
 
 	// ingress controller configuration options
 	ingressControllerDisabled bool
+	ingressControllerImage    string
+	ingressControllerImageTag string
 
 	// proxy server general configuration options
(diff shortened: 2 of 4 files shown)


shaneutt (changes requested):
Looks good!

My only comment is about tests. We should add (or update) an integration test that verifies this flag is working by using it (defining a specific older version of the ingress controller) and then verifying that when the environment is created the correct image was applied to the `Deployment` and that the environment deploys successfully. This will help protect us against regressions.

As for testing the CLI side of things, we haven't historically had automated tests for much of that. We don't need to balloon the scope of your PR to include starting a test suite for the CLI, but if you felt strongly that you _wanted_ to do it by all means go for it. Otherwise just make sure you've tested it manually locally before we merge.

jrsmroz:
I've tested that manually. I'll take a look at the integration tests and see what can be done.

shaneutt:
> I've tested that manually. I'll take a look at the integration tests and see what can be done.

Sounds good :+1: 

This test is pretty straightforward and shows the basic workflow of setting up addons (in your case, adding a builder option for the image version), building the environment, and then testing to make sure things are working as expected: https://github.com/Kong/kubernetes-testing-framework/blob/main/test/integration/metallb_test.go#L21

pkg/clusters/addons/kong/builder.go
```diff
@@ -113,6 +117,14 @@ func (b *Builder) WithProxyImage(repo, tag string) *Builder {
 	return b
 }
 
+// WithControllerImage configures the ingress controller
+// container image name and tag for the Kong proxy.
```
pmalek, line 121: I believe this will be a bit more clear for the reader. WDYT?

```suggestion
// WithControllerImage configures the ingress controller container image name and tag.
```

internal/cmd/ktf/environments.go
```diff
@@ -210,14 +211,28 @@ func configureKongAddon(cmd *cobra.Command, envBuilder *environments.Builder) *e
...
+
+	if customControllerImage != "" {
+		imageParts := strings.Split(customControllerImage, ":")
+		if len(imageParts) == 1 {
+			imageParts = append(imageParts, "latest")
+		}
+		if len(imageParts) != 2 { //nolint:gomnd
+			cobra.CheckErr(fmt.Errorf("malformed --kong-ingress-controller-image: %s", customControllerImage))
+		}
+		builder.WithControllerImage(imageParts[0], imageParts[1])
```
pmalek, line 233: Well this doesn't look strictly right. If `len(imageParts) == 1` then we'll also get the error right? Maybe we should put this section in a `switch/case` statement?

jrsmroz: If len is 1 we append `latest` and len will be 2.

pmalek: Ah ok.

And then what happened is that it wasn't obvious to me that `cobra.CheckErr()` calls `os.Exit()`, hence I thought the program will continue with `len(imageParts) != 2`.

So this works but it's not easy to comprehend at first glance :) I'm ok with leaving this as is though 👍

pmalek (approved):

pmalek:
BTW: in `configureKongAddon()` we're checking the `kong-disable-controller` flag. Wouldn't it make more sense to make setting the image conditional on this flag being unset?

jrsmroz:
It makes sense. Just I tried to make it same as setting `proxyImage`, which does not check for the `ingressControllerDisabled` flag. In the end I don't think  that matters, because if `ingressControllerDisabled` the controller will not be enabled.

Setting the proxy image was not tested too. I'm adding tests for both options.

test/integration/kongaddon_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	t.Log("verifying that the kong deployment is using custom images")
+	deployments := env.Cluster().Client().AppsV1().Deployments(kong.Namespace())
+	kongDeployment, err := deployments.Get(ctx, "ingress-controller-kong", metav1.GetOptions{})
+	require.NoError(t, err)
+	require.Len(t, kongDeployment.Spec.Template.Spec.Containers, 2)
+	require.Equal(t, kongDeployment.Spec.Template.Spec.Containers[0].Name, "ingress-controller")
+	require.Equal(t, kongDeployment.Spec.Template.Spec.Containers[1].Name, "proxy")
+	require.Contains(t, kongDeployment.Spec.Template.Spec.Containers[0].Image, tc.controllerImage())
+	require.Contains(t, kongDeployment.Spec.Template.Spec.Containers[1].Image, tc.proxyImage())
```
shaneutt, line 81: Kind of a minor thing, but we would expect these to actually be equal, would we not?

```suggestion
	require.Equal(t, kongDeployment.Spec.Template.Spec.Containers[0].Image, tc.controllerImage())
	require.Equal(t, kongDeployment.Spec.Template.Spec.Containers[1].Image, tc.proxyImage())
```

shaneutt (approved):
Looking good, just one last minor comment before merging.

> Setting the proxy image was not tested too. I'm adding tests for both options.

Also wanted to thank you for taking the time to add that: there's probably unfortunately a few things that don't have proper tests but now that we know there are a few different end-users it's becoming more and more important to get the test coverage up, so much appreciated. :vulcan_salute:

test/integration/kongaddon_test.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (tc customImageTest) name() string {
+	return fmt.Sprintf("KongAddonImages:[%s,%s]", tc.controllerImage(), tc.proxyImage())
+}
+
+func TestKongAddontWithCustomImage(t *testing.T) {
+	tests := []customImageTest{
+		{"kong/kubernetes-ingress-controller", "2.3.0", "kong", "2.7"},
+		{"kong/kubernetes-ingress-controller", "2.3.1", "kong", "2.8"},
```
pmalek, line 39: Would it be controversial from my side to suggest adding field names in place where struct initialization happens? This way it'd be obvious which fields are initialized with which values (without looking at struct definition)

test/integration/kongaddon_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (tc customImageTest) proxyImage() string {
+	return fmt.Sprintf("%s:%s", tc.proxyImageRepo, tc.proxyImageTag)
+}
+
+func (tc customImageTest) name() string {
+	return fmt.Sprintf("KongAddonImages:[%s,%s]", tc.controllerImage(), tc.proxyImage())
+}
+
+func TestKongAddontWithCustomImage(t *testing.T) {
```
pmalek, line 36:
```suggestion
func TestKongAddonWithCustomImage(t *testing.T) {
```

test/integration/kongaddon_test.go
```diff
@@ -0,0 +1,82 @@
...
+	tests := []customImageTest{
+		{"kong/kubernetes-ingress-controller", "2.3.0", "kong", "2.7"},
+		{"kong/kubernetes-ingress-controller", "2.3.1", "kong", "2.8"},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name(), func(t *testing.T) {
+			t.Parallel()
+			testKongAddonWithCustomImage(t, tc)
```
pmalek, line 45: Any particular reason for extracting this to a separate function? I do agree with making the functions short and succinct but I wonder what's the value of putting the whole test body in a separate `func`

jrsmroz: Just a personal preference to keep those as separate functions.

pmalek: +1 for striving for small indentation, I'm not sure I see the benefits of using the separate functions like that.

Maybe I'll get convinced with time 🤷‍♂️

pmalek (approved):

PR #309, opened 2022-07-01
feat: add Kuma addon

rainest:
Add a Kuma addon. Following the recent ServiceAccount token integration breakage, we need to add Kuma to integration tests. First go is a bare-bones addon that can deploy an MTLS mesh policy and not much else.

We'll probably be a quite limited if we want to do anything with this beyond basic smoke testing: Kuma apparently lacks a clientset, so we either need to contribute that or build everything from YAML. That works in Helm tests but is iffy in the Golang tests.

Example usage in this KIC draft: https://github.com/Kong/kubernetes-ingress-controller/pull/2635

### DIFF
--- a/pkg/clusters/addons/kuma/addon.go
+++ b/pkg/clusters/addons/kuma/addon.go
@@ -1,0 +1,237 @@
+package kuma
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"os"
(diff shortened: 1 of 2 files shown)


pmalek (approved):

pmalek:
Please address the failed lint. Other than that: 🚢

PR #314, opened 2022-07-08
feat: add support for Kong pull secret

rainest:
Adds Kong Builder options for creating and using a Docker pull Secret.


pkg/clusters/addons/kong/addon.go
```diff
@@ -190,6 +198,36 @@ func (a *Addon) Deploy(ctx context.Context, cluster clusters.Cluster) error {
...
+		cmd = exec.CommandContext(ctx, "kubectl", args...)
+		cmd.Stdout = io.Discard
+		cmd.Stderr = stderr
+		if err := cmd.Run(); err != nil {
+			return fmt.Errorf("%s: %w", stderr.String(), err)
+		}
+		// use the pull Secret
+		a.deployArgs = append(a.deployArgs, []string{
+			"--set", "image.pullSecrets={proxy-pull}",
+		}...)
```
shaneutt, line 228: Ideally we should be moving in a direction of reducing `kubectl` calls to make the framework more versatile. Since we're in a position to use `cluster.Client()` here to create the secret I think we should do so, rather than adding more `kubectl` calls.

shaneutt (changes requested):

pkg/clusters/addons/kong/addon.go
```diff
@@ -190,6 +199,36 @@ func (a *Addon) Deploy(ctx context.Context, cluster clusters.Cluster) error {
...
+			Name:     "proxy-pull",
+			Username: a.proxyPullSecret.Username,
+			Email:    a.proxyPullSecret.Email,
+			Password: a.proxyPullSecret.Password,
+			Server:   a.proxyPullSecret.Server,
+		}
+		if opts.Server == "" {
+			opts.Server = "https://index.docker.io/v1/"
+		}
+		secret := opts.createSecretDockerRegistry()
```
shaneutt, line 220: This is a private method?

shaneutt (approved):
Honestly I couldn't help but laugh that we replaced `kubectl` calls with calls to `k8s.io/kubectl` :joy: but it's part of their public API so I don't really have any reason to oppose it. LGTM.

pkg/clusters/addons/kong/addon.go
rainest: There is no escape from kubectl, only exec, and unfortunately it's not particularly amenable to re-using its helpers and requires everything effectively run as normal kubectl minus cobra :shrug:

rainest:
Post-review push added the test and fixed some garbage necessary to play nice with the kubectl `Run()` invocation.

shaneutt (approved):

rainest:
Additional follow-up push was forgetting to tidy mod, so not dismissing for that one.

shaneutt (approved):

jrsmroz (approved):

jrsmroz:
Tests times out after 45 minutes 🤔

shaneutt:
> Tests times out after 45 minutes thinking

During that same timeframe I saw kind based integration tests failing over multiple repositories, I think something was going on with GH actions :thinking: but github status isn't reporting anything :eyes:

rainest:
TestEnvironmentWithKnative never completed in the bad run, but ran fine locally. Observed the same weird timeouts in the KIC repo, so looks like maybe a runner fluke.

PR #326, opened 2022-07-14
chore: bump GKE test version

rainest:
https://github.com/Kong/kubernetes-testing-framework/runs/7345390116?check_suite_focus=true failed because GKE 1.19 is no longer available. Bumped to latest version on the regular channel.


pmalek (approved):
I wonder if there's an API that can return an array of actually supported versions 🤔

PR #334, opened 2022-07-28
Improve Kong diagnostic config dumps

rainest:
Update the Kong addon's diagnostics to:
- Extract DB-less config YAML from the mangled admin API JSON response.
- Add support for DB-backed config using decK parts.

Added 0.17 changelog since I'd like to try and figure out what the heck is going on with the frequent nightly Kong test failures and the tests that flake on their own, and I assume we want the CNI stuff available for the operator.

[ktf-diag-2409850005.tar.gz](https://github.com/Kong/kubernetes-testing-framework/files/9214720/ktf-diag-2409850005.tar.gz) is example Postgres (Enterprise) output.
[ktf-diag-2834659457.tar.gz](https://github.com/Kong/kubernetes-testing-framework/files/9214773/ktf-diag-2834659457.tar.gz) shows the updated DB-less output.


CHANGELOG.md
```diff
@@ -17,6 +17,10 @@
 - Added `ApplyManifestByURL` and `DeleteManifestByURL` helper functions to the
   `cluster` package as siblings to `ApplyYAML` and `DeleteYAML`.
   [#330](https://github.com/Kong/kubernetes-testing-framework/pull/330)
+- Added a diagnostics system to collect resources, describe information, pod
+  logs, and available plugin diagnostics.
+  [#332](https://github.com/Kong/kubernetes-testing-framework/pull/332)
+  [#334](https://github.com/Kong/kubernetes-testing-framework/pull/334)
```
shaneutt, line 23: Needs to be moved to v0.17.1 or v0.18.0 at your discretion.

shaneutt (commented):
It occurs to me that we're missing tests for this stuff entirely. Can we add some basic tests so that we can help prevent regressions where we might break the dumps in the future?

rainest:
871f7f7 piggybacks on this since it's hopefully small enough and avoids waiting for the changelog update in a separate PR. While converting some of the problem KIC tests to use this system, I realized that we couldn't get the correct defer ordering for examples tests, since those need to delete YAML instead of resources, and the cleaner lacked support for that.

pkg/clusters/addons/kong/addon.go
```diff
@@ -404,7 +485,16 @@ func (a *Addon) DumpDiagnostics(ctx context.Context, cluster clusters.Cluster) (
 		if err != nil {
 			return diagnostics, err
 		}
-		diagnostics["dbless_config.json"] = b.Bytes()
+		var kongConfig struct {
+			Config string `json:"config,omitempty" yaml:"config,omitempty"`
+		}
+		err = json.Unmarshal(b.Bytes(), &kongConfig)
```
pmalek, line 491: I think you drop all of those `bytes.Buffer` shenanigans and just use: https://pkg.go.dev/encoding/json#Decoder.Decode like so

```go
json.NewDecoder(resp.Body).Decode(&kongConfig)
```

pkg/clusters/addons/kong/addon.go
```diff
@@ -394,6 +402,79 @@ func (a *Addon) DumpDiagnostics(ctx context.Context, cluster clusters.Cluster) (
...
+		}
+		client, err := deckutils.GetKongClient(opts)
+		if err != nil {
+			return diagnostics, fmt.Errorf("could not build Kong client: %w", err)
+		}
+		workspaces, err := client.Workspaces.ListAll(ctx)
+		var kongAPIError *kong.APIError
+		if errors.As(err, &kongAPIError) && kongAPIError.Code() == http.StatusNotFound {
+			defaultws := kong.Workspace{Name: kong.String("default")}
+			workspaces = append(workspaces, &defaultws)
```
pmalek, line 433: Am I getting this right that this relies on the fact that if the `err` is non `nil` then  `workspaces` are empty hence we can `append` on it and get a slice with 1 element? If so I'd urge us no to do so, but just assign `workspaces` to a slice with one element: `defaultws`. WDYT?

shaneutt (approved):

Issue #331, opened 2022-07-25
Need a method to test/wait for service to be access inside cluster


randmonkey:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem Statement

When we deploy some addons or services, we need a tool to detect or wait for the service to be ready to serve. For example, when we deploy `kuma` addon in the cluster, we need to wait for kuma's webhook service to be ready before we create CRDs in kuma. So, we need a method to test whether a service could be accessed inside the testing kubernetes cluster. 

### Proposed Solution

- provide a solution that runs a `Job` with some images to test the TCP connection periodically, and complete the job when the TCP connection is created.

### Acceptance Criteria

 could call a method in KTF to test if a service/address could be accessed in the test cluster by:
  - [ ] (At least) TCP connection
  - [ ] GET/POST to an HTTP URL
  - [ ] TLS connection (with CA/client cert)
  - [ ] GET/POST by an HTTPS URL (with a CA/client cert)

mflendrich:
Reason: Fixing this test flake https://github.com/Kong/kubernetes-ingress-controller/issues/2735 requires that KTF waits for Redis availability.

randmonkey:
test flakiness that could be resolved or debugged by this issue:
- https://github.com/Kong/kubernetes-ingress-controller/issues/2733
- https://github.com/Kong/kubernetes-ingress-controller/issues/2735
- https://github.com/Kong/gateway-operator/issues/100

randmonkey:
Proposed solution:
- create an image `tcptest` that exits with code 0 when successfully connected to a TCP address, and retry until a configurable timeout, and exit with non-0 code when reached timeout
- when testing a service, create a `Job` using this image and wait for completion of that job

rainest:
Kubernetes should (effectively) provide this for us, since we can check Deployment status. This isn't directly linked to the Service status (though you can figure it out by querying labels), but that's usually a non-issue since we just know the correct Deployment in our tests.

For https://github.com/Kong/gateway-operator/issues/100 we are checking the Deployment readiness, but are still failing. Why the Deployment is ready without the webhook is unclear, since we do add a readiness requirement using controller-runtime and use its standard webhook server tooling, which I'd hope starts the server before saying it's ready.

 For https://github.com/Kong/kubernetes-ingress-controller/issues/2733 we are not checking the Deployment readiness or retrying the apply yet. This may be affected by the same Deployment-ready-webhook-not-ready problem as the above.

[...]

PR #336, opened 2022-08-01
feature: add job to test connectivity to tcp service

randmonkey:
Try to add a job to test connectivity to a TCP service inside a k8s cluster to solve #331.
DONE in this PR:
- create a image retries to connect to a TCP address and push it to Kong repo
- KTF provide a method to wait for a TCP address to be connected by the image above

TODO:
- support more parameters (connect timeout/retry interval or retry times)
- discuss about other kinds of "tests", http/https/tls/udp ?

### DIFF
--- a/.github/workflows/build-push-tcptest-image.yaml
+++ b/.github/workflows/build-push-tcptest-image.yaml
@@ -1,0 +1,92 @@
+name: build and push tcptest image
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'The version to release (e.g. v0.18.0)'
+        required: true
+      latest:
+        description: 'Whether to tag this release latest'
+        required: true
+        default: 'false'
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Parse semver string
+        id: semver_parser
+        uses: booxmedialtd/ws-action-parse-semver@v1
+        with:
+          input_string: ${{ github.event.inputs.tag }}
+          version_extractor_regex: 'v(.*)$'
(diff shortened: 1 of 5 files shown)


shaneutt (changes requested):
I'm really not sure about this approach. Last we had talked I thought we were going to add a tool that deploys a temporary `LoadBalancer` type `Service` to the cluster, so that in this way we can just test using the Go library and not need a new image and all this other stuff.

What was wrong with the `LoadBalancer` approach?

randmonkey:
> I'm really not sure about this approach. Last we had talked I thought we were going to add a tool that deploys a temporary `LoadBalancer` type `Service` to the cluster, so that in this way we can just test using the Go library and not need a new image and all this other stuff.
> 
> What was wrong with the `LoadBalancer` approach?

The reason why I proposed this approach (OR, the cons for `LoadBalancer` approach): 
- `LoadBalancer` may not be supported by all kinds of clusters. If a cluster does not provide a `LoadBalancer` service, the temp lb solution could not be used.
- The new temporary service is not exactly the same thing as the original one. So the result may not be 100% prove that the original service could/could not be accessed. 

Cons for this approach:
 - We need to maintain extra code and image to deploy the jobs inside cluster.

The `LoadaBalancer` solution is here: #338. We could consider which solution would be accepted.

shaneutt:
In a testing environment we _control the cluster environment_ and can guarantee `LoadBalancer` functionality is available, and additionally that deals with the problem utilizing Kubernetes native APIs that can be injected into any test rather than a job that needs to be maintained as a separate tool. I am highly in favor of us taking the `LoadBalancer` approach instead for these reasons.

PR #338, opened 2022-08-03
feat: wait for service connective by TCP

### LINKED ISSUE #331: Need a method to test/wait for service to be access inside cluster

randmonkey:
Implement An approach to test TCP connectivity of a service by creating a temp LB service. fixes #331.

### DIFF
--- a/pkg/utils/kubernetes/networking/services.go
+++ b/pkg/utils/kubernetes/networking/services.go
@@ -3,7 +3,11 @@ package networking
 import (
 	"context"
 	"fmt"
+	"net"
+	"time"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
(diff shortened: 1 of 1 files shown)


pkg/utils/kubernetes/networking/services.go
```diff
@@ -45,3 +48,92 @@ func WaitForServiceLoadBalancerAddress(ctx context.Context, c *kubernetes.Client
 		}
 	}
 }
+
+// WaitForServicePortConnected waits for a service provided by namespace/name
+// being able to connect by TCP. It creates a LoadBalancer with same selector and ports,
+// and try to establish TCP connection to the loadbalacer IP and port.
```
shaneutt, line 54:
```suggestion
// WaitForServicePortConnected waits until it can make successful TCP connections
// to a service (provided by namespace/name). This will temporarily create a LoadBalancer
// type Service to allow connections to the Service and port from outside the cluster while
// the connection attempts are made using the LoadBalancer public address.
```

pkg/utils/kubernetes/networking/services.go
```diff
@@ -45,3 +48,92 @@ func WaitForServiceLoadBalancerAddress(ctx context.Context, c *kubernetes.Client
...
+func WaitForServicePortConnected(ctx context.Context, c *kubernetes.Clientset, namespace, name string, port int) error {
```
shaneutt, line 55: What do you think about this alternative name suggestion?

```suggestion
func WaitForConnectionOnServicePort(ctx context.Context, c *kubernetes.Clientset, namespace, name string, port int) error {
```

pkg/utils/kubernetes/networking/services.go
```diff
@@ -45,3 +48,92 @@ func WaitForServiceLoadBalancerAddress(ctx context.Context, c *kubernetes.Client
...
+	}
+
+	// empty selector, we should create the endpoints separately.
+	if service.Spec.Selector == nil || len(service.Spec.Selector) == 0 {
+		endpoints, err := c.CoreV1().Endpoints(namespace).Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		// copy the endpoints for the lb service, and set metadata.
+		tempEndPioints := endpoints.DeepCopy()
```
shaneutt, line 85:
```suggestion
		tempEndPoints := endpoints.DeepCopy()
```

pkg/utils/kubernetes/networking/services.go
```diff
@@ -45,3 +48,92 @@ func WaitForServiceLoadBalancerAddress(ctx context.Context, c *kubernetes.Client
...
+		tempEndPioints.ObjectMeta = metav1.ObjectMeta{
+			Namespace: namespace,
+			Name:      lbServiceName,
+			Labels: map[string]string{
+				"corresponding-service": name,
+			},
+		}
+		_, err = c.CoreV1().Endpoints(namespace).Create(ctx, tempEndPioints, metav1.CreateOptions{})
```
shaneutt, line 93: I feel like I might be missing something simple, can you help me to better understand why we're _creating_ endpoints?

pkg/utils/kubernetes/networking/services.go
```diff
@@ -45,3 +48,92 @@ func WaitForServiceLoadBalancerAddress(ctx context.Context, c *kubernetes.Client
...
+	}
+
+	ticker := time.NewTicker(time.Second)
+	address := fmt.Sprintf("%s:%d", ip, port)
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("context completed while waiting for %s:%d to be connected", ip, port)
+		case <-ticker.C:
+			// TODO: add a timeout for connecting to the address and make it configurable?
```
shaneutt, line 132: Yes, we should use a timeout for this. You can find some examples of this in the [KIC TCPRoute Tests](https://github.com/Kong/kubernetes-ingress-controller/blob/main/test/integration/tcproute_test.go)

shaneutt (changes requested):
Mostly minor comments, overall looks good :+1:

pkg/utils/kubernetes/networking/services.go
randmonkey: updated.

pkg/utils/kubernetes/networking/services.go
randmonkey: updated.

pkg/utils/kubernetes/networking/services.go
randmonkey: fixed.

pkg/utils/kubernetes/networking/services.go
randmonkey: some services may not have an effective `spec.selector`. Their upstreams are specified by an `endpoints` resource with the same name created separately (for example, service for validating webhook in KIC integration tests, which pointing to the webhook server run outside testing k8s cluster). So we should also create endpoints here with same configuration of target service. 
I explained this in the comments, please give more comments if they did not express well.

pkg/utils/kubernetes/networking/services.go
randmonkey: updated, added a `dialTimeout` parameter.

shaneutt (approved):

PR #341, opened 2022-08-04
fix: increase retry timeout to 1m on creating mesh in kuma

randmonkey:
increase retry times to wait more time (1m) for Kuma webhook startup. Try to solve [KIC#2733](https://github.com/Kong/kubernetes-ingress-controller/issues/2733).

### DIFF
--- a/pkg/clusters/addons/kuma/addon.go
+++ b/pkg/clusters/addons/kuma/addon.go
@@ -226,7 +226,7 @@ spec:
 // enableMTLS attempts to apply a Mesh resource with a basic retry mechanism to deal with delays in the Kuma webhook
 // startup
 func (a *Addon) enableMTLS(ctx context.Context, cluster clusters.Cluster) (err error) {
-	for i := 0; i < 5; i++ {
+	for i := 0; i < 12; i++ {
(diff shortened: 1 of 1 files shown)


pkg/clusters/addons/kuma/addon.go
```diff
@@ -226,7 +226,7 @@ spec:
 // enableMTLS attempts to apply a Mesh resource with a basic retry mechanism to deal with delays in the Kuma webhook
 // startup
 func (a *Addon) enableMTLS(ctx context.Context, cluster clusters.Cluster) (err error) {
-	for i := 0; i < 5; i++ {
+	for i := 0; i < 12; i++ {
```
pmalek, line 229: It just occurred to me that this code block might be better off when implemented with a `select` on `ctx.Done()` and `time.After()` channels so that it can be interrupted, contrary to `time.Sleep()`.

pmalek (approved):
Approved but please consider my comment in refactoring this into a `select`.

pkg/clusters/addons/kuma/addon.go
randmonkey: Yes, we may need change this part to use `select` and `time.After`.

PR #342, opened 2022-08-05
chore: refactor enableMTLS in kuma addon and update CHANGELOG.md

randmonkey:
- refactor `enableMTLS` in kuma addon to use `select` for retrying
- update CHANGELOG.md for releasing v0.19.0 of KTF

### DIFF
--- a/pkg/clusters/addons/kuma/addon.go
+++ b/pkg/clusters/addons/kuma/addon.go
@@ -226,13 +226,21 @@ spec:
 // enableMTLS attempts to apply a Mesh resource with a basic retry mechanism to deal with delays in the Kuma webhook
 // startup
 func (a *Addon) enableMTLS(ctx context.Context, cluster clusters.Cluster) (err error) {
-	for i := 0; i < 12; i++ {
-		err = clusters.ApplyManifestByYAML(ctx, cluster, mtlsEnabledDefaultMesh)
-		if err != nil {
-			time.Sleep(time.Second * 5) //nolint:gomnd
-		} else {
-			break
+	ticker := time.NewTicker(5 * time.Second) // nolint:gomnd
+	timeoutTimer := time.NewTimer(time.Minute)
+
+	for {
+		select {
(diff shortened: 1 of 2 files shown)


pkg/clusters/addons/kuma/addon.go
```diff
@@ -226,13 +226,23 @@ spec:
 // enableMTLS attempts to apply a Mesh resource with a basic retry mechanism to deal with delays in the Kuma webhook
 // startup
 func (a *Addon) enableMTLS(ctx context.Context, cluster clusters.Cluster) (err error) {
-	for i := 0; i < 12; i++ {
-		err = clusters.ApplyManifestByYAML(ctx, cluster, mtlsEnabledDefaultMesh)
-		if err != nil {
-			time.Sleep(time.Second * 5) //nolint:gomnd
-		} else {
-			break
+	timer := time.NewTicker(5 * time.Second) // nolint:gomnd
```
pmalek, line 229:
```suggestion
	ticker := time.NewTicker(5 * time.Second) // nolint:gomnd
```

pkg/clusters/addons/kuma/addon.go
```diff
@@ -226,13 +226,23 @@ spec:
...
+	retryTimes := 0
+	maxRetries := 12
```
pmalek, line 231: What you could consider, refactoring this further, is to add a timer that's set to hit after a designated wait time instead relying on an implicit `maxRetries=12` and `tick` duration being 5 seconds.

This way you'll get more (arguably) idiomatic code with another select case below. 

WDYT?

pkg/clusters/addons/kuma/addon.go
randmonkey: fixed.

pkg/clusters/addons/kuma/addon.go
randmonkey: updated.

pmalek (approved):

PR #343, opened 2022-08-05
fix: use correct command for describe

rainest:
Didn't change the command after copy pasting originally, whoops.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,15 +3,22 @@
 ## v0.19.0 
 
 ### Added
+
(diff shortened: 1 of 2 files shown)


pmalek (approved):

CHANGELOG.md
```diff
@@ -3,16 +3,23 @@
 ## v0.19.0
```
pmalek, line 3: I have a feeling we're a little bit too liberal with changes to changelog, namely the version/tag should only be added when we release. Before that the newest entry in the changelog should always be "unreleased".

PR #395, opened 2022-09-26
fix: move error log and release 0.22.1

rainest:
metallb IPAddressPool creation failures are currently logging lines like:

```
failed to deploy addon metallb: failed to create metallb.io/v1beta1 IPAddressPool: context deadline exceeded, last error on create: Post "https://127.0.0.1:36091/apis/metallb.io/v1beta1/namespaces/metallb-system/ipaddresspools": context deadline exceeded
```
Code review suggests that since we're assigning lastErr after the creation attempt but before the switch handling the `ctx.Done()` case, where the switch context is the same context we pass to the create request, the last create request fails because the context is done and then wipes out the previous error. That previous error was hopefully more interesting, because it should be whatever prevented the loop from completing before the context expired.

Moving the lastErr update inside the context-not-expired switch case should avoid this and thus log the last not-context expired error. Not tested though since I don't have a quick way to trigger the situation. Please double-check my thinking here.

Release a patch release for this fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## v0.22.1
+
+### Fixed
+
+- Moved metallb error recording inside context switch, to avoid
+  recording the context expiration as the last error.
+
 ## v0.22.0
 
 ### Added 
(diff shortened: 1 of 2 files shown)


jrsmroz (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## v0.22.1
+
+### Fixed
+
+- Moved metallb error recording inside context switch, to avoid
```
pmalek, line 7: add pr link?

rainest: Gimme some hook magic that automatically replaces a template string with the PR number. We're kinda lax on it, especially for single-change releases. Ultimately blame can usually tell you, unless someone else added the entry retroactively.

PR #400, opened 2022-09-29
fix: ignore metallb admission

rainest:
Set the metallb admission controller to ignore. Our IPAddressPools and L2Advertisements are static and work. The admission controller has a habit of not coming online and blocking the deploy. Ignoring its failures avoids this.

Add a kustomize utility and refactor the metallb deploy to use it.

The kustomize utility is generic enough to go into KTF. The specific patches and Kustomization structs will be provided by individual tests (or, in KTF's case, plugins). This is adapted from the code originally introduced in https://github.com/Kong/kubernetes-ingress-controller/pull/2995 and would replace it in KIC once released. It differs in that it instead takes a variadic manifest argument and appends these to the resources, rather than replacing the resources entirely with a single manifest resource. This allows for additional resources as well as URLs (URLs you can just stick in the Kustomization directly; they don't need a helper to write the local file and inject a generated filename).

### DIFF
--- a/pkg/utils/kubernetes/kubectl/kustomize.go
+++ b/pkg/utils/kubernetes/kubectl/kustomize.go
@@ -1,0 +1,58 @@
+package kubectl
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/ghodss/yaml"
+	"sigs.k8s.io/kustomize/api/krusty"
+	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+// GetKustomizedManifest takes a kustomization and any number of manifest readers. It adds the manifests to the
+// kustomization's resources and returns a reader with the rendered kustomization.
+func GetKustomizedManifest(kustomization types.Kustomization, manifests ...io.Reader) (io.Reader, error) {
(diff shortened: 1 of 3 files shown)


pkg/utils/kubernetes/kubectl/kustomize.go
```diff
@@ -0,0 +1,58 @@
...
+	for i, manifest := range manifests {
+		orig, err := io.ReadAll(manifest)
+		if err != nil {
+			return nil, err
+		}
+		err = os.WriteFile(filepath.Join(workDir, fmt.Sprintf("base_%d.yaml", i)), orig, 0o600) //nolint:gomnd
+		if err != nil {
+			return nil, err
+		}
+		kustomization.Resources = append(kustomization.Bases, fmt.Sprintf("base_%d.yaml", i))
```
pmalek, line 33: 1. Why assign to `kustomization.Resources` but append to `kustomization.Bases`?
2. `bases` was deprecated in v2.1.0 https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/bases/

rainest: It was originally written using the latter (a lot of the Kustomize docs examples still use bases) and my brain is conflating the two :shrug:

jrsmroz (approved):

pmalek (approved):

PR #415, opened 2022-10-17
refactor: move `DumpDiagnostics` from `Cleaner` to `Cluster`

pmalek:
This PR moves the `DumpDiagnostics` from `Cleaner` to `Cluster`.

This will reflect the responsibility ownership a bit more accurately as the cleaner should have nothing to do with the diagnostics data gathering: this proves to be the fact since the actual implementation of `DumpDiagnostics` does not rely on any other field of the `Cleaner` struct other than the `Cluster`.

Additionally to the above, specific cluster implementations can now extend (or redefine) the diagnostics gathering to perhaps gather data specific for them.


pkg/clusters/cleanup.go
```diff
@@ -50,169 +45,6 @@ func (c *Cleaner) AddNamespace(namespace *corev1.Namespace) {
 	c.namespaces = append(c.namespaces, namespace)
 }
 
-// DumpDiagnostics gathers a wide range of diagnostic information from the test cluster, to provide a snapshot of it
-// at a given time for offline debugging. It uses the provided context and writes the meta string to meta.txt to
-// identify the result set.
-func (c *Cleaner) DumpDiagnostics(ctx context.Context, meta string) (string, error) {
```
czeslavo, line 56: WDYT about keeping it for compatibility? It could just delegate the call to the `c.cluster`.

pmalek: Good idea. I've added a wrapper that delegates the call to the underlying cluster with a deprecated note. PTAL.

czeslavo (approved):

PR #416, opened 2022-10-17
tests: use Github Actions matrix for integration tests

pmalek:
This PR makes it so that ktf integration tests are using a matrix on CI now in such a way that each test runs in it's own workflow. This way we don't expect one test disrupting another one or problems with performance (since default, hosted Github runners are generally known of their not so great performance).

Additionally, this PR adds `integration-tests-passed` job, which has been made required from now on. This makes this setup "scalable" in a way such that when new integration tests are added no actions have to be done in order to make the CI happy.

This also shuffles around the environments to prevent spam in Github conversation view: where not needed the `environment: "integration-tests"` setting has been removed and an additional `setup-kong-license` job has been added to pull the license in 1 place.

Last but not least, this passes the kong license as an encrypted output between `setup-kong-license` and `integration-tests` so that it doesn't get blocked by Github on suspicion of passing secrets/masked variables between actions. The implemented solution was based on https://nitratine.net/blog/post/how-to-pass-secrets-between-runners-in-github-actions/

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -14,10 +14,6 @@ jobs:
     environment: "integration-tests"
     runs-on: ubuntu-latest
     steps:
-    - uses: Kong/kong-license@master
-      id: license
-      with:
-        password: ${{ secrets.PULP_PASSWORD }}
 
     - name: setup golang
       uses: actions/setup-go@v3
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -3,14 +3,14 @@ name: tests
 on:
   pull_request:
     branches:
-      - 'main'
+    - 'main'
   push:
     branches:
-      - 'main'
+    - 'main'
   workflow_dispatch: {}
 
 jobs:
-  tests-and-coverage:
+  unit-tests:
     environment: "integration-tests"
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yaml
```diff
@@ -40,14 +40,70 @@ jobs:
...
+      test_names: ${{ steps.set_test_names.outputs.test_names }}
+    steps:
+    - uses: actions/checkout@v3
+    - id: set_test_names
+      name: Set test names
+      working-directory: test/integration/
+      # grep magic described in https://unix.stackexchange.com/a/13472
+      # sed to add the extra $ is because some of our test names overlap. we need it so the -run regex only matches one test
+      run: |
+        echo "::set-output name=test_names::$(grep -shoP "(?<=^func )(Test[a-zA-z_0-9]+)(?=\(t \*testing.T\) {)" * | sed -e "s/$/\$/"| jq -R . | jq -cs .)"
```
jrsmroz, line 63: Knowing that `set-output` is to be deprecated, can we make it already use env file?

pmalek: You're right. I was just lazy I guess. PTAL now.

jrsmroz (approved):
lgtm

PR #459, opened 2022-12-06
chore(deps): use only sigs.k8s.io/yaml to handle yaml

mmorel-35:


go.mod
```diff
@@ -7,46 +7,33 @@ require (
 	github.com/blang/semver/v4 v4.0.0
 	github.com/docker/docker v20.10.21+incompatible
 	github.com/google/go-github/v39 v39.2.0
+	github.com/google/go-github/v48 v48.1.0
```
mmorel-35, line 10: It's seems you are using two versions of go-github module

PR #464, opened 2022-12-07
chore(deps): use only github.com/google/go-github last version

mmorel-35:
And also apply go mod tidy


rainest:
Not sure if we recently broke the license handling for this repo or it never worked. Merging and watching the tests on main after, since it's unlikely the library change broke anything if it's passing vet and unit tests.

rainest:
https://github.com/Kong/kubernetes-testing-framework/actions/runs/3644364486/jobs/6153536793

PR #467, opened 2022-12-08
chore(deps): bump golang.org/x/oauth2 to v0.3.0

mmorel-35:


rainest:
Continuing in https://github.com/Kong/kubernetes-testing-framework/pull/487 to avoid the CI limitations on external contribs. We should probably improve those for this library; it shouldn't need Enterprise for _everything_.

PR #487, opened 2023-01-05
Bump OAuth2 lib

rainest:
Redo of https://github.com/Kong/kubernetes-testing-framework/pull/467 to avoid external secret limitations. KTF doesn't like those.


pmalek (approved):
Can you be more descriptive on why we need this? (besides just bumping the dependency)

rainest:
No reason in particular, just a routine version bump.

I think Dependabot wasn't getting it because something had previously pinned it to a SHA instead of a tag? Not sure, but I can't see any other reason it wouldn't get bumped automatically.

rainest:
Heck, they even had another release since the original suggested bump. Onwards to 0.4.0!

pmalek (approved):

PR #498, opened 2023-01-18
refactor(gke): use grpc when createSubnet set

czeslavo:
When reading the documentation again, I realized that the `--create-subnet` can be used in the gRPC API: https://cloud.google.com/kubernetes-engine/docs/how-to/alias-ips#api_1.


pmalek (approved):
Nice 🎖️

PR #510, opened 2023-01-25
fix: fix Cleaner's cleanup for Gateway API resources

pmalek:
For some reason `.TypeMeta.Kind` and `.TypeMeta.APIVersion` are empty when we're working with Gateway API objects.

This causes the [`Delete()` in the `Cleaner.Cleanup()`](https://github.com/Kong/kubernetes-testing-framework/blob/48d69a4fe8ac235d2de5b53e0dac5a0a103b6510/pkg/clusters/cleanup.go#L63) to always return a 404 while the object is still on the cluster.

This PR adds some code that manually fixes this by amending Gateway API objects' `Kinds` and `APIVersion` fields.

It also adds a check to prevent reporting not found errors during namespace deletion: no point in reporting an error if it's already deleted right?


pkg/clusters/cleanup.go
```diff
@@ -114,6 +117,88 @@ func (c *Cleaner) Cleanup(ctx context.Context) error {
 	return g.Wait()
 }
 
+// fixupObjKinds takes a client.Object and checks if it's of one of the gateway
+// API types and if so then it adjusts that object's Kind and APIVersion.
+// The reason we need this is that when interacting via Gateway API objects
+// for some reason .TypeMeta.Kind and .TypeMeta.APIVersion are empty.
```
czeslavo, line 123: Are we sure that's behavior specific to Gateway API objects only? 🤔 What would be the reason for that?

czeslavo (dismissed):

pkg/clusters/cleanup.go
pmalek: I was afraid I'm gonna fall into a rabbit whole of sorts so I just left it at that :)

czeslavo: Alright. I was just curious if that could be done in a more generic way with a clever dose of reflection. Let's not fall into that hole for now. 👍

pmalek: I've added links to issues in kubernetes that I believe are related.

czeslavo (approved):

PR #516, opened 2023-01-31
ci: use gh in install.sh for artifacts download

pmalek:
This PR aims to avoid hitting the API rate limiting by using `gh` which should already be using `GITHUB_TOKEN` on Github's CI.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -1,5 +1,11 @@
 name: tests
 
+concurrency:
+  # Run only for most recent commit in PRs but for all tags and commits on main
+  # Ref: https://docs.github.com/en/actions/using-jobs/using-concurrency
+  group: ${{ github.workflow }}-${{ github.head_ref || github.sha }}
+  cancel-in-progress: true
+
 on:
   pull_request:
     branches:
@@ -10,6 +16,21 @@ on:
   workflow_dispatch: {}
 
 jobs:
+  installer-tests:
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: checkout repository
+      uses: actions/checkout@v3
+
+    - name: run installer script
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
+      run: ./docs/install.sh
+
+    - name: run ktf
+      run: ~/.local/bin/ktf --help
+
   unit-tests:
     runs-on: ubuntu-latest
     steps:
--- a/docs/install.sh
+++ b/docs/install.sh
@@ -10,8 +10,7 @@ BASE="kong/kubernetes-testing-framework"
 
 echo "INFO: verifying operating system compatibility"
 
-OSTYPE=$(uname -s)
-OSTYPE=$(echo "${OSTYPE,,}") # convert to lower case
+OSTYPE=$(uname -s | tr '[:upper:]' '[:lower:]')
 
 if [ "$OSTYPE" != "linux" ]; then
     if [ "$OSTYPE" != "darwin" ]; then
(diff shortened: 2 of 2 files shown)


docs/install.sh
```diff
@@ -26,8 +25,7 @@ fi
 
 echo "INFO: verifying cpu architecture compatibility"
 
-ARCH=$(uname -m)
-ARCH=$(echo "${ARCH,,}") # convert to lower case
+ARCH=$(uname -m | tr '[:upper:]' '[:lower:]')
```
shaneutt, line 28: Were these changed because they were causing some problem? `tr` is a GNU coreutil, is it present on a Mac system (I don't use Mac so I know very little about them except that they're originally FreeBSD kernel based)?

pmalek: It didn't work for me on Mac - probably because it has version 3.2 - (while `tr` does) nor on a VM with BusyBox (since it uses `ash`).

shaneutt (dismissed):
Let's just make sure we add helpful output if a user uses this script but doesn't have `gh` installed to go install it, and then test it on our supported platforms, then LGTM :+1:

pmalek:
> Let's just make sure we add helpful output if a user uses this script but doesn't have `gh` installed to go install it, and then test it on our supported platforms, then LGTM 👍

Thanks for the suggestion: I ended up testing this on Mac arm64 and Linux aarch64 (as reported by `uname`) alpine image. This needed some changes.

PTAL.

shaneutt (dismissed):
Looks OK, just as long as we test before merging.

I would also ask if we haven't already, we should create some automated tests for this installer and to exercise the CLI to ensure we're not breaking the CLI and its installer on updates. If we could at minimum create a follow-up issue for that, and manually test now that would be good :+1:

pmalek:
I've added a very simple installer test: example run: https://github.com/Kong/kubernetes-testing-framework/actions/runs/4076445875/jobs/7024196104#step:3:10

Plus: as noted above I've tested it manually which should do for now. The old behaviour of downloading via `curl` stays, so there shouldn't ™️ be any surprises there.

PTAL.

.github/workflows/tests.yaml
```diff
@@ -1,5 +1,11 @@
 name: tests
 
+concurrency:
+  # Run only for most recent commit in PRs but for all tags and commits on main
+  # Ref: https://docs.github.com/en/actions/using-jobs/using-concurrency
+  group: ${{ github.workflow }}-${{ github.head_ref || github.sha }}
+  cancel-in-progress: true
```
pmalek, line 7: Adding this to prevent simultaneous runs for the same SHA

pmalek:
Force merging because @shaneutt is not a member of team required to approve PRs in this repo 🤷

.github/workflows/tests.yaml
```diff
@@ -10,6 +16,21 @@ on:
...
+    - name: checkout repository
+      uses: actions/checkout@v3
+
+    - name: run installer script
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
+      run: ./docs/install.sh
+
+    - name: run ktf
+      run: ~/.local/bin/ktf --help
```
mflendrich, line 32: What is the purpose of `--help` here? 🤔

if the purpose is to validate that the ktf binary is present, please (at least) comment about the intent behind using `--help` in an automated workflow

docs/install.sh
```diff
@@ -26,44 +25,67 @@ fi
...
 
 if [ "$ARCH" = x86_64 ]; then
     ARCH="amd64"
 fi
+if [ "$ARCH" = aarch64 ]; then
+    ARCH="arm64"
+fi
 
-if [ "$ARCH" != "amd64" ]; then
+if [[ "$ARCH" != "amd64" && "$ARCH" != "arm64" ]]; then
```
mflendrich, line 37: nit: inconsistent quotes around arch names `"arm64"` vs `aarch64` in lines 37 and 33 respectively

docs/install.sh
```diff
@@ -26,44 +25,67 @@ fi
...
     echo "Error: ${ARCH} is not a supported architecture at this time."
     exit 1
 fi
 
-# ------------------------------------------------------------------------------
-# Determine Latest Release
-# ------------------------------------------------------------------------------
+TEMPDIR=$(mktemp -d)
```
mflendrich, line 42: can we delete the tempdir when done with it? maybe something along the lines of `trap rm -rv "$TEMPDIR" EXIT` ?

docs/install.sh
```diff
@@ -26,44 +25,67 @@ fi
...
-if [ "$LATEST_RELEASE" = "" ]; then
-    echo "Error: could not find latest release for ${BASE}!${LATEST_RELEASE}"
-    exit 1
-fi
+    # ------------------------------------------------------------------------------
+    # Determine Latest Release
+    # ------------------------------------------------------------------------------
 
-echo "INFO: the latest release of ${BASE} was determined to be ${LATEST_RELEASE}"
+    LATEST_RELEASE=$(curl -s https://api.github.com/repos/${BASE}/releases/latest | perl -ne 'print $1 if m{"name": "(v[0-9]+\.[0-9]+\.[0-9]+.*)"}')
```
mflendrich, line 52: there's already a pattern for tool download in the Makefile:


https://github.com/Kong/kubernetes-ingress-controller/blob/4187223cfffce38fb0a6691a565534a0176f8711/Makefile#L34

docs/install.sh
pmalek: It's being done in lines 113-115. Not via a `trap` (which would be better) but it's there.

pmalek: Created an issue for this: #531

docs/install.sh
pmalek: It is indeed. But the purpose of this PR was not to change this mechanism but rather introduce a new way of downloading via `gh`: which download release artifacts, not builds from source via `go install ...`.

We might think about this as a fallback but then what would be the point of this script if it was an equivalent of calling `go install github.com/kong/kubernetes-testing-framework/cmd/ktf@latest` 🤔

PR #525, opened 2023-02-01
fix(ci): fix release names

pmalek:
- No need to specify the release and tag names, otherwise they'll have `refs/tags/` prefix
- Don't mark releases as drafts or prereleases

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -102,12 +102,9 @@ jobs:
 
     - name: Release
       uses: softprops/action-gh-release@v1
-      id: create_release
       with:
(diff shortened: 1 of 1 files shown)


rainest (approved):
Any reason to remove the ID? I guess we're not using it, so it doesn't matter, but it seems an odd thing to strip out.

pmalek:
After I shuffled stuff around in https://github.com/Kong/kubernetes-testing-framework/commit/2bbe4667b958ff9448275cd47704ba66a67d813f I kept it and didn't notice it wasn't necessary anymore hence remove it in here.

PR #527, opened 2023-02-02
fix: fix command error handling when error is nil

pmalek:


shaneutt (approved):
There's no description, but this seems to be self-explanatory. LGTM.

PR #532, opened 2023-02-03
chore: add changelog for v0.28.1

pmalek:


czeslavo (approved):

pmalek:
> I imagine there's some way we can configure Codecov to not track the CHANGELOG as something that needs to be tested? 🤔

More CI magic, `if` this `then` that sort of thing. I recall we already tried it somewhere.

The problem we hit was that when you have a required check on github you have to somehow code around it when you don't run tests that make this check pass.

PR #549, opened 2023-02-07
ci: fix condition for running e2es

czeslavo:
We should be explicit about the value we check as for some reason when the result of the first step was 'false', we still got other steps run: https://github.com/Kong/kubernetes-testing-framework/actions/runs/4117158754/jobs/7108115410.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -153,25 +153,25 @@ jobs:
         run: echo "result=${{ secrets.PULP_PASSWORD != '' && secrets.GOOGLE_APPLICATION_CREDENTIALS != '' }}" >> $GITHUB_OUTPUT
 
       - uses: Kong/kong-license@master
-        if: steps.detect_if_should_run.outputs.result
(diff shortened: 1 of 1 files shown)


pmalek (approved):
Makes sense given that we always set `result` 🤔 👍

czeslavo:
> Makes sense given that we always set `result` 🤔 👍

I'm afraid we might have the same issue in https://github.com/Kong/kubernetes-ingress-controller/blob/main/.github/workflows/test.yaml that hasn't exhibited itself yet?

pmalek:
> > Makes sense given that we always set `result` 🤔 👍
> 
> I'm afraid we might have the same issue in https://github.com/Kong/kubernetes-ingress-controller/blob/main/.github/workflows/test.yaml that hasn't exhibited itself yet?

Where specifically ? https://github.com/Kong/kubernetes-ingress-controller/blob/1084eb60d52bd89d7f852abce70d186e8bb0562c/.github/workflows/test.yaml#L159-L161 ?

PR #552, opened 2023-02-10
feat: make cleaner safe for concurrent use

pmalek:


randmonkey (dismissed):

czeslavo (approved):

pkg/clusters/cleanup.go
```diff
@@ -40,18 +42,26 @@ func NewCleaner(cluster Cluster) *Cleaner {
...
 
 func (c *Cleaner) AddNamespace(namespace *corev1.Namespace) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
 	c.namespaces = append(c.namespaces, namespace)
 }
 
 func (c *Cleaner) Cleanup(ctx context.Context) error {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
```
shaneutt, line 64: I'm not sure in practice if this will end up mattering, but it seems a bit odd to make the actual cleanup a "ready only" lock, considering all the mutations it's making.

Actually, it occurs to me now that what we should probably do is lock the object and block all future changes once we've run cleanup. E.g. any `Add` to the cleaner fails after `Cleanup` is called.

pmalek: I've create an issue to track this: https://github.com/Kong/kubernetes-testing-framework/issues/563.

This should be OK for now (at least for our internal usage). We might take another approach at this when indeed we use cleaners more than once.

PR #553, opened 2023-02-10
fix: fix endless loop when cleaning up namespaces

pmalek:


pmalek:
Created #554 to address vague error printed in https://github.com/Kong/kubernetes-testing-framework/actions/runs/4143881679/jobs/7166276368

```
=== RUN   TestKindClusterWithCalicoCNI
=== PAUSE TestKindClusterWithCalicoCNI
=== CONT  TestKindClusterWithCalicoCNI
    calico_test.go:28: configuring the test environment with Calico enabled
    calico_test.go:31: building the testing environment and Kubernetes cluster
    calico_test.go:33: 
        	Error Trace:	/home/runner/work/kubernetes-testing-framework/kubernetes-testing-framework/test/integration/calico_test.go:33
        	Error:      	Received unexpected error:
        	            	exit status 1
        	Test:       	TestKindClusterWithCalicoCNI
--- FAIL: TestKindClusterWithCalicoCNI (36.45s)
```

randmonkey (approved):

pkg/clusters/cleanup.go
```diff
@@ -84,7 +84,12 @@ func (c *Cleaner) Cleanup(ctx context.Context) error {
...
 
 			if err := namespaceClient.Delete(ctx, namespace.Name, metav1.DeleteOptions{}); err != nil {
-				if !errors.IsNotFound(err) {
+				switch errors.IsNotFound(err) {
+				// If the namespace cannot be found then we're good to go.
+				case true:
+					return nil
+				// If there was a different error then report it to the caller.
+				case false:
 					return err
```
pmalek, line 93:
```suggestion
				if errors.IsNotFound(err) {
					// If the namespace cannot be found then we're good to go.
					return nil
				}
				return err
```

Let's get back to `if` 😅 . WDYT?

shaneutt (commented):
My only would-be comment would be partially a nitpick that for a two case switch statement where the expression is simply a boolean, I would normally just use an `if/else`, which I think we should change to.  Otherwise though, looks like this will fix the issue thanks for taking care of it :+1:

czeslavo (dismissed):

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Fix an endless loop in cleaner's `Cleanup()` when a namespace to be deleted
+  is alread gone.
```
pmalek, line 6:
```suggestion
  is already gone.
```

czeslavo (approved):

pkg/clusters/cleanup.go
```diff
@@ -94,8 +94,11 @@ func (c *Cleaner) Cleanup(ctx context.Context) error {
 			namespaceClient := c.cluster.Client().CoreV1().Namespaces()
 
 			if err := namespaceClient.Delete(ctx, namespace.Name, metav1.DeleteOptions{}); err != nil {
-				if !errors.IsNotFound(err) {
-					return err
+				if errors.IsNotFound(err) {
+					// If the namespace cannot be found then we're good to go.
+					return nil
+				}
+				return err
```
shaneutt, line 101: extra bracket below here

pkg/clusters/cleanup.go
```diff
@@ -94,8 +94,11 @@ func (c *Cleaner) Cleanup(ctx context.Context) error {
...
 				}
 			}
```
pmalek, line 103:
```suggestion
				if errors.IsNotFound(err) {
					// If the namespace cannot be found then we're good to go.
					return nil
				}
				return err
			}
```

pkg/clusters/cleanup.go
pmalek: Yup, that's what happens when you code via github code suggestions.

shaneutt (approved):

PR #567, opened 2023-02-15
feat: add machine type param to gke cluster builder

pmalek:
Related: https://github.com/Kong/kubernetes-ingress-controller/issues/3555

### DIFF
--- a/pkg/clusters/types/gke/builder.go
+++ b/pkg/clusters/types/gke/builder.go
@@ -26,21 +26,27 @@ type Builder struct {
 	jsonCreds         []byte
 	waitForTeardown   bool
 
-	createSubnet   bool
-	addons         clusters.Addons
-	clusterVersion *semver.Version
-	majorMinor     string
-	labels         map[string]string
+	createSubnet    bool
+	addons          clusters.Addons
+	clusterVersion  *semver.Version
+	majorMinor      string
+	nodeMachineType string
+	labels          map[string]string
 }
+
+const (
+	defaultNodeMachineType = "e2-highcpu-4"
+)
 
 // NewBuilder provides a new *Builder object.
 func NewBuilder(gkeJSONCredentials []byte, project, location string) *Builder {
 	return &Builder{
-		Name:      fmt.Sprintf("t-%s", uuid.NewString()),
-		project:   project,
-		location:  location,
-		jsonCreds: gkeJSONCredentials,
-		addons:    make(clusters.Addons),
+		Name:            fmt.Sprintf("t-%s", uuid.NewString()),
+		project:         project,
+		location:        location,
+		jsonCreds:       gkeJSONCredentials,
+		nodeMachineType: defaultNodeMachineType,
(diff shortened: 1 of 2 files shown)


pkg/clusters/types/gke/builder.go
```diff
@@ -26,21 +26,27 @@ type Builder struct {
...
+	createSubnet    bool
+	addons          clusters.Addons
+	clusterVersion  *semver.Version
+	majorMinor      string
+	nodeMachineType string
+	labels          map[string]string
 }
 
+const (
+	defaultNodeMachineType = "e2-highcpu-4"
```
shaneutt, line 38: Not a blocker but I'm curious: is this something derivable from the googlecloud packages we import?

pkg/clusters/types/gke/builder.go
```diff
@@ -26,21 +26,27 @@ type Builder struct {
...
+
 // NewBuilder provides a new *Builder object.
 func NewBuilder(gkeJSONCredentials []byte, project, location string) *Builder {
 	return &Builder{
-		Name:      fmt.Sprintf("t-%s", uuid.NewString()),
-		project:   project,
-		location:  location,
-		jsonCreds: gkeJSONCredentials,
-		addons:    make(clusters.Addons),
+		Name:            fmt.Sprintf("t-%s", uuid.NewString()),
```
shaneutt, line 44: Not a blocker: It occurs to me that this _might_ be a good time to improve the names of these clusters, like even just doing this I think makes it very clear (to us) at a glance in the google cloud dashboard what kinds of clusters these are:

```suggestion
		Name:            fmt.Sprintf("ktf-%s", uuid.NewString()),
```

pkg/clusters/types/gke/builder.go
pmalek: A quick code search on Github and through https://grep.app/ doesn't show any exposed `const`s that we could use.

pkg/clusters/types/gke/builder.go
pmalek: I'm OK with this change and it makes sense to me but I wonder if we wouldn't have propblems with cluster name limits? I'm not even sure what is the limit as of now but I'd prefer to check this beforehand.

shaneutt: Yeah I'm not sure. I'll consider this resolved, and at your discretion if you find that it wont be a problem and you think it's worthwhile, go for it. This is a non-blocker.

czeslavo: Yeah, I remember trying to change this naming pattern to something more direct instead of just a `t-` prefix and GCP rejected it due to names being too long. Anyways, it's quite easy to tell that the cluster was created by KTF as we add labels to tell that.

czeslavo (approved):

pmalek:
@czeslavo PTAL again. I forgot that rebasing with Github UI doesn't sign the commits 🤦

czeslavo (approved):

PR #603, opened 2023-04-05
chore(ci): retry codecov upload if it fails

pmalek:
To avoid spurious failures in codecov report upload like:

```
[2023-04-05T09:20:58.253Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 502 - 
<html><head>
<meta http-equiv="content-type" content="text/html;charset=utf-8">
<title>502 Server Error</title>
</head>
<body text=#000000 bgcolor=#ffffff>
<h1>Error: Server Error</h1>
<h2>The server encountered a temporary error and could not complete your request.<p>Please try again in 30 seconds.</h2>
<h2></h2>
</body></html>
```

https://github.com/Kong/kubernetes-testing-framework/actions/runs/4616788061/jobs/8162299292#step:7:103

, retry the upload 3 times instead of failing the CI.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -48,14 +48,23 @@ jobs:
     - name: run unit tests
       run: make test.unit
 
+    # We're using a retry mechanism for codecov to ensure we do get the reports
+    # uploaded. The alternative is to use fail_ci_if_error: false, but that
+    # somewhat defeats the purpose of uploading those reports. Why bother uploading
+    # if we don't care if the upload's successful?
     - name: Upload coverage to Codecov
-      uses: codecov/codecov-action@v3
+      if: steps.detect_if_should_run.outputs.result == 'true'
+      uses: Wandalen/wretry.action@v1.0.36
       with:
-        token: ${{ secrets.CODECOV_TOKEN }}
-        fail_ci_if_error: true
-        flags: unit-test
-        files: unit.coverage.out
-        verbose: true
+        action: codecov/codecov-action@v3
+        with: |
+          token: ${{ secrets.CODECOV_TOKEN }}
+          fail_ci_if_error: true
+          flags: unit-test
+          files: unit.coverage.out
+          verbose: true
+        attempt_limit: 3
+        attempt_delay: 30000
 
   setup-integration-tests:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -48,14 +48,20 @@ jobs:
...
       run: make test.unit
 
+
     - name: Upload coverage to Codecov
-      uses: codecov/codecov-action@v3
+      if: steps.detect_if_should_run.outputs.result == 'true'
+      uses: Wandalen/wretry.action@v1.0.36
       with:
-        token: ${{ secrets.CODECOV_TOKEN }}
-        fail_ci_if_error: true
```
czeslavo, line 55: Alternatively, we could use `fail_ci_if_error: false` and accept that it may in some cases fail. <controversy> Codecov is not that critical in the end to make it a required check. </controversy> WDYT? I'd prefer that as it's just simpler.

pmalek: I thought about this too but then what's the point of those reports if we don't care about whether they are uploaded or not? Either commit to having them or just straight up abort uploading them.

czeslavo: Yeah, I get your point. Let's maybe just add a comment above the step to explain why we want to retry codecov action? Noting that "given it relies on an external service that has proven to temporarily fail, we don't want it to affect the results of CI" would suffice.

go.mod
```diff
@@ -24,9 +24,9 @@ require (
 	k8s.io/api v0.26.3
 	k8s.io/apiextensions-apiserver v0.26.3
 	k8s.io/apimachinery v0.26.3
-	k8s.io/cli-runtime v0.26.2
+	k8s.io/cli-runtime v0.26.3
 	k8s.io/client-go v0.26.3
-	k8s.io/kubectl v0.26.2
+	k8s.io/kubectl v0.26.3
```
czeslavo, line 29: Are these bumps expected in this PR?

pmalek: Nope, good catch. PTAL

.github/workflows/tests.yaml
pmalek: Added the comment. PTAL

czeslavo (approved):

PR #613, opened 2023-04-07
chore(ci): bump codecov retry limit to 10

pmalek:
It seems that there's been occasions where the retry helped (e.g. here https://github.com/Kong/kubernetes-testing-framework/actions/runs/4630922396/jobs/8193283484?pr=609 2 failed attempts and the third being successful) whereas sometimes even 3 retries is not enough (e.g. here https://github.com/Kong/kubernetes-testing-framework/actions/runs/4630922396/jobs/8193283631?pr=609)

To try to address this let's bump the limit of retries to 10.


shaneutt (approved):
Curious about _why_ it needs to retry but since a part of this is somewhat out of our control and if we see that a couple more retries might help, sure.

pmalek:
> Curious about _why_ it needs to retry but since a part of this is somewhat out of our control and if we see that a couple more retries might help, sure.

It seems that it's more than just us experiencing this issue: https://github.com/codecov/codecov-action/issues/926

PR #630, opened 2023-04-19
feat: dump logs by `kind export` in kind clusters

randmonkey:
As described in #591, switch the DumpDiagnostics function to use command `kind export` for cluster log dumping instead. This command will also dump the logs of kubelet, which would help when failed to create kubernetes cluster.
The original PR removed dumping of `kubectl get all` and `kubectl describe all`, I added them back in the [commit](https://github.com/Kong/kubernetes-testing-framework/pull/630/commits/255b67ff20c7849cd1e549cb6a890641662cf73d).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+- Dump pod logs by `kind export logs` in kind clusters. This will also dump
+  logs of kubernetes components not running in pods, including kubelet, 
+  containerd, etc.
+  [#630](https://github.com/Kong/kubernetes-testing-framework/pull/630)
+  [#591](https://github.com/Kong/kubernetes-testing-framework/pull/591)
+
 ## v0.30.1
 
 - Upgrade `metallb` addon to `v0.13.9`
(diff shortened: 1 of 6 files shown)


pmalek:
@randmonkey Is the aim of this to replace what #591 is proposing?

randmonkey:
> @randmonkey Is the aim of this to replace what #591 is proposing?

Yes, the origin intention was to dump k8s component logs when we failed to wait for env ready, which may caused by failure of creating k8s cluster.

shaneutt (changes requested):
As I understand it this fixes a problem that was described in feedback for #591. That PR is still in progress and the author has been in touch with us so it's not clear to me what the impetus for steamrolling that PR is: if getting this functionality is time-sensitive for other projects, we could have (and should have) explained that to the original author that we had a time crunch and needed to push things through faster, but as far as I can see we didn't do that here. Our new contributor wasn't given proper time to address the feedback yet so if possible we should provide more suggestions (ideally, Github suggestions which are easily actionable) related to the changes we're hoping for, OR perhaps we could suggest the changes as a follow-up and allow that PR to merge first and then build on top of it.

PR #632, opened 2023-04-21
test: skip cases requiring license in integration tests and update GH workflow

randmonkey:
partial of #596, enable forks to run integration tests not requiring an enterprise license.
Note: This PR is used to test the effect of running integration tests on forks.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -96,9 +96,14 @@ jobs:
     steps:
     # This step is needed to avoid running the integration tests requiring an enterprise license
     # if the secrets are not available.
-    - name: Detect if we should run and set env to enable tests requring a license (have required secrets)
+    - name: Detect if we should run test cases requring an enterprise license (have required secrets)
       id: detect_if_should_run_enterprise
-      run: echo "result=${{ secrets.PULP_PASSWORD != '' }}" >> $GITHUB_OUTPUT && echo "KTF_TEST_RUN_ENTERPRISE_CASES=true" >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yaml
```diff
@@ -94,32 +94,34 @@ jobs:
...
+    # This step is needed to avoid running the integration tests requiring an enterprise license
+    # if the secrets are not available.
+    - name: Detect if we should run test cases requring an enterprise license (have required secrets)
+      id: detect_if_should_run_enterprise
       run: echo "result=${{ secrets.PULP_PASSWORD != '' }}" >> $GITHUB_OUTPUT
 
+    - name: Set environment variable to enable test cases requiring an enterprise license
+      if: steps.detect_if_should_run_enterprise.outputs.result == 'true'
+      id: set_run_enterprise_env
+      run: echo "KTF_TEST_RUN_ENTERPRISE_CASES=true" >> $GITHUB_ENV
```
pmalek, line 106: Is there a reason we need to introduce a new environment variable instead of using `KONG_LICENSE_DATA` https://github.com/Kong/kubernetes-testing-framework/blob/51dc612f3da531099d8ff3105d0018a9447b4c22/pkg/clusters/addons/kong/enterprise.go#L21 ?

test/integration/enterprise_test.go
```diff
@@ -25,6 +25,7 @@ import (
 )
 
 func TestKongEnterprisePostgres(t *testing.T) {
+	SkipEnterpriseTestIfNoEnv(t)
```
pmalek, line 28: Can we rename this to explicitly say that we're looking for a license not just "an env"?

pmalek (approved):

PR #643, opened 2023-05-07
feat(k8s): Switch from CoreV1 Endpoints to DiscoveryV1 EndpointSlice

programmer04:
EndpointSlice API is available as V1 since K8s 1.21. Let's do not rely on old Endpoint API in KTF.

It's been spotted during work on https://github.com/Kong/kubernetes-ingress-controller/issues/3916.
[`TestValidationWebhook`](https://github.com/Kong/kubernetes-ingress-controller/blob/aad1dbb1513dd49a14db0f3dc8b6ad4158c2a292/test/integration/webhook_test.go#LL43C6-L43C27) from KIC [uses `WaitForConnectionOnServicePort` function](https://github.com/Kong/kubernetes-ingress-controller/blob/aad1dbb1513dd49a14db0f3dc8b6ad4158c2a292/test/integration/webhook_test.go#L682) from KTF. When KIC drops `Endpoints` in favour of `EndpointSlice` this test doesn't pass, because `Endpoints` are not created (and KTF expecting them).

Mirroring works in other way from `Endpoints` to `EndpointSlices` [under circumstances described in docs](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#endpointslice-mirroring). Thus after this change KTF should be still backward compatible with `Endpoints` and does not break someone's tests that rely on old `Endpoints`.

I've tested this change with KIC and everything passes
- KIC with old `Endpoints` and KTF with `EndpointSlice`
- KIC with `EndpointSlice` (WIP) and KTF with `EndpointSlice`

Thus this PR with high certainty (maybe some edge cases exist) shouldn't break someone's setup. 

Other improvements - reduce coupling by replacing concrete K8s client as parameter with corresponding interface.

### DIFF
--- a/pkg/utils/kubernetes/networking/services.go
+++ b/pkg/utils/kubernetes/networking/services.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	discoveryv1 "k8s.io/api/discovery/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
@@ -17,7 +18,7 @@ import (
 // address. This function will throw an error if the service gets provisioned
 // more than a single address, that is not supported. The context provided
 // should have a timeout associated with it or you're going to have a bad time.
-func WaitForServiceLoadBalancerAddress(ctx context.Context, c *kubernetes.Clientset, namespace, name string) (string, bool, error) {
+func WaitForServiceLoadBalancerAddress(ctx context.Context, c kubernetes.Interface, namespace, name string) (string, bool, error) {
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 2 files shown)


pkg/utils/kubernetes/networking/services.go
```diff
@@ -54,57 +55,66 @@ func WaitForServiceLoadBalancerAddress(ctx context.Context, c *kubernetes.Client
...
-			if err != nil && !errors.IsNotFound(err) {
-				fmt.Printf("failed to delete endpoints %s/%s after testing, error %v\n",
-					namespace, lbServiceName, err)
+			for _, eps := range tempEndpointSlices {
+				err := c.DiscoveryV1().EndpointSlices(namespace).Delete(ctx, eps.Name, metav1.DeleteOptions{})
+				if err != nil && !errors.IsNotFound(err) {
+					fmt.Printf("failed to delete endpoints %s/%s after testing, error %v\n",
+						namespace, eps.Name, err,
+					)
+				}
```
pmalek, line 117: Shouldn't that whole block land just after the `Create()` above? This way if `i`th `EndpointSlice` where `i < n` and `n` is the length of the `tempEndpointSlices` returns an error upon creation you get the other ones - that were already created - removed.

programmer04: Very, good catch! Thanks, I proposed a solution that ensures every successfully created temporary EndpointSlice will be clean up on return. For each one deffered clean-up function is created, in Go [stack is huge](https://dave.cheney.net/2013/06/02/why-is-a-goroutines-stack-infinite) and I think performance penalty for having multiple `defer` will be negligible in this case

pkg/utils/kubernetes/networking/services.go
```diff
@@ -54,66 +55,72 @@ func WaitForServiceLoadBalancerAddress(ctx context.Context, c *kubernetes.Client
...
+			epsName := fmt.Sprintf("%s-%d", lbServiceName, i)
+			tempEndpointSlices[i].ObjectMeta = metav1.ObjectMeta{
+				Namespace: namespace,
+				Name:      epsName,
+				Labels: map[string]string{
+					discoveryv1.LabelServiceName: lbServiceName, // Maps EndpointSlice to Service.
+					correspondingSvcNameLabel:    name,
+				},
+			}
+			if _, err = c.DiscoveryV1().EndpointSlices(namespace).Create(ctx, &tempEndpointSlices[i], metav1.CreateOptions{}); err != nil {
```
pmalek, line 105: 👍 

one nit: you might want to extract `c.DiscoveryV1().EndpointSlices(namespace)` to a named variable this way making the call sites more readable.

pkg/utils/kubernetes/networking/services.go
```diff
@@ -54,66 +55,72 @@ func WaitForServiceLoadBalancerAddress(ctx context.Context, c *kubernetes.Client
...
+				return err
 			}
-		}()
+			// For each successfully created temporary EndpointSlice ensure deletion on return from the function.
+			defer func(namespace, epsName string) {
+				err := c.DiscoveryV1().EndpointSlices(namespace).Delete(ctx, epsName, metav1.DeleteOptions{})
+				if err != nil && !errors.IsNotFound(err) {
+					fmt.Printf("failed to delete EndpointSlice %s/%s after testing, error %v\n",
+						namespace, epsName, err,
+					)
```
pmalek, line 114: This might be refactored into returning this error in the enclosing `func` so that it's not lost among the logged messages.

We might create an issue for this for later, to solve it in a separate PR.

pkg/utils/kubernetes/networking/services.go
programmer04: good advice, done

pkg/utils/kubernetes/networking/services.go
programmer04: Yeah, let's create an issue. I think the whole KTF should have some more standardised logging / error returning strategy

pmalek: While at it: do you mind creating an issue to test `WaitForConnectionOnServicePort()`? We don't want to rely on KIC's tests to inform us about failures in its implementation.

pmalek (approved):
Ideally we'd work on the tests now - #649 - to ensure the changed functions still work 😅

pmalek:
@programmer04 You'll need to re-push the commits after you've signed them

PR #646, opened 2023-05-08
chore: document knative.Builder.WithVersion

czeslavo:
Document the expected version format passed to `knative.Builder.WithVersion`.

### DIFF
--- a/pkg/clusters/addons/knative/builder.go
+++ b/pkg/clusters/addons/knative/builder.go
@@ -14,7 +14,7 @@ func NewBuilder() *Builder {
 	return &Builder{version: DefaultVersion}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/clusters/addons/knative/builder.go
```diff
@@ -1,31 +1,32 @@
 package knative
 
 import (
-	"errors"
+    "errors"
```
pmalek, line 4: These spaces don't look right. Should be tabs for *.go files

czeslavo: For some reason, my IDE (GoLand) has detected this project uses spaces 🤔 Sorry for that.

pmalek (approved):

PR #704, opened 2023-06-12
ci: fix integration tests aggregate check

pmalek:
The existing `integration-tests-passed` condition passes even when some of the integration tests failed, e.g. https://github.com/Kong/kubernetes-testing-framework/actions/runs/5241672836/jobs/9464102490?pr=703.

This PR tries to address this.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -150,7 +150,7 @@ jobs:
 
   integration-tests-passed:
     needs: integration-tests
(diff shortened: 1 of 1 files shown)


programmer04 (approved):

pmalek:
This seems to be working. Failing `TestKongWithNodePort` indeed blocks the CI. #705 aims to fix that test and after pulling those changing here we should be good to go.

PR #737, opened 2023-07-06
feat: dump diagnostics if cluster not ready

rainest:
If the cluster fails to become ready after 20 minutes, assume it's stuck and dump a diagnostic.

At present unready clusters will just hit the test timeout and exit without providing any useful information.

### DIFF
--- a/pkg/clusters/types/gke/cluster.go
+++ b/pkg/clusters/types/gke/cluster.go
@@ -230,7 +230,7 @@ func (c *Cluster) DumpDiagnostics(ctx context.Context, meta string) (string, err
 	defer os.Remove(kubeconfig.Name())
 
 	// create a tempdir
-	outDir, err := os.MkdirTemp(os.TempDir(), "ktf-diag-")
+	outDir, err := os.MkdirTemp(os.TempDir(), clusters.DiagnosticOutDirectoryPrefix)
 	if err != nil {
 		return "", err
 	}
--- a/pkg/environments/implementation.go
+++ b/pkg/environments/implementation.go
@@ -16,7 +16,11 @@ import (
 // Test Environment - Implementation
 // -----------------------------------------------------------------------------
 
-const objectWaitSleepTime = time.Millisecond * 200
+const (
+	objectWaitSleepTime = time.Millisecond * 200
+
+	readyHungDuration = time.Minute * 20
+)
 
 // environment is the default KTF Environment used for testing Kubernetes ingress.
 type environment struct {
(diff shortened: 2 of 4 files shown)


pkg/clusters/types/gke/cluster.go
```diff
@@ -230,7 +230,7 @@ func (c *Cluster) DumpDiagnostics(ctx context.Context, meta string) (string, err
 	defer os.Remove(kubeconfig.Name())
 
 	// create a tempdir
-	outDir, err := os.MkdirTemp(os.TempDir(), "ktf-diag-")
+	outDir, err := os.MkdirTemp(os.TempDir(), clusters.DiagnosticOutDirectoryPrefix)
```
shaneutt, line 233: :+1:

shaneutt (dismissed):
Simple and clear, looks good. The only thing that gave me any pause was the exact duration of `readyHungDuration`, is 20 minutes _too_ much time? I guess it will vary from system to system and it's really hard to predict how Github actions runners will perform any given day, so yeah I can see this being a reasonable starting point and we can tune it further if we find any need. :+1:

pkg/environments/implementation.go
```diff
@@ -81,11 +85,22 @@ func (env *environment) WaitForReady(ctx context.Context) chan error {
 	errs := make(chan error)
 
 	go func() {
+		// if the cluster fails to become ready after N minutes, assume it's likely stuck and dump a diagnostic bundle.
+		// this uses its own timer since we can't catch "go test" timeouts via the ctx.
+		hung := time.AfterFunc(readyHungDuration, func() {
```
pmalek, line 90: We might want to check if context has deadline via `ctx.Deadline()` and if it does use it here. Otherwise fallback to a const.

This can be handy in tests using `t.Deadline()` which could then be passed to context to `WaitForReady()`

WDYT?

rainest: If we remember to use it, yes.

I think we just collect on any done condition then, deadline or otherwise, since it expiring here should mean _something_ unusual happened. Will maybe generate diagnostics if you terminate the process here, but whatever, tmpfiles are cheap.

rainest:
20m is indeed arbitrary based on not much more than our having a 45m timeout. If you use something shorter, passing the deadline to the context is the proper way of doing it.

randmonkey:
> 20m is indeed arbitrary based on not much more than our having a 45m timeout. If you use something shorter, passing the deadline to the context is the proper way of doing it.

So could we set this timeout to a variable and make it adjustable by env variables?

pmalek:
> > 20m is indeed arbitrary based on not much more than our having a 45m timeout. If you use something shorter, passing the deadline to the context is the proper way of doing it.
> 
> So could we set this timeout to a variable and make it adjustable by env variables?

I'd advocate for using test's deadline, this way we control it with only 1 variable : test timeout.

pmalek (approved):

PR #738, opened 2023-07-06
feat: upgrade metallb implemenation for 0.15.x

rainest:
Use the metallb implementation from 0.31, as the old implementation no longer works.

To handle backports in KIC 2.5, we need to continue using KTF 0.15. KTF includes dependencies on client-go, GWAPI, etc., and upgrading those by pulling in the new KTF breaks API compatibility for 2.5. However, the 0.15 metallb addon [no longer works](https://github.com/Kong/kubernetes-ingress-controller/pull/4281/commits/bcd09158dff12a4e6e66dd817e2104a0c63547ab).

This change pulls in the latest addon in its entirety and creating a release branch in prep to release a 0.15.2. This technically breaks semver because it's new functionality (although it does not break API compatibility), but I'm not sure what we can really do otherwise since we can't exactly bump the minor version (0.16 already exists). We may want to consider cutting new KTF major releases when we want to do a KIC LTS, although this tightly couples KTF version bumps to something it's not directly.


pmalek (approved):
> This technically breaks semver because it's new functionality

We're pre v1 (do what you want/need) so that doesn't break semver.

pmalek:
The CI is not functional though 😢

pmalek (approved):

PR #752, opened 2023-07-22
feat: add IPv6-only clusters

rainest:
Support spawning KIND as IPv6 only.

No planned GKE support for this at present. It looks like GKE _can_ now do IPv6, but setting it up requires building a VPC? It's something more complicated than just a cluster setup toggle.

Adds IPv6 support for metallb and kong plugins. The latter requires chart 2.27+, but doesn't bump requirements for non-IPv6 functionality.

I'm not entirely sure what the intent of the old metallb range selection was, since it apparently just chose a static set of arbitrary start and end IPs within the upper 2/3 of a /24. The test netmask was also non-canonical, which causes weirdness. I think the new approach of choosing the actual upper half of the given network is close enough.


test/integration/ipv6_test.go
```diff
@@ -0,0 +1,67 @@
...
+		if net.ParseIP(service.Status.LoadBalancer.Ingress[0].IP).To4() != nil {
+			return false
+		}
+		kongURL := fmt.Sprintf("http://[%s]", service.Status.LoadBalancer.Ingress[0].IP)
+		resp, err := http.Get(kongURL)
+		// we don't care that the proxy has nothing to serve so long as we can talk to it and get a valid HTTP response
+		if err == nil && resp.StatusCode == http.StatusNotFound {
+			return true
+		}
+		return false
```
czeslavo, line 64: Wouldn't checking the status code to be 404 be a potential for flakiness?
```suggestion
		_, err = http.Get(kongURL)
		if err != nil {
			return false
		}
		// We only care if we can talk to the proxy and get a valid HTTP response.
		return true
```

czeslavo (approved):

test/integration/ipv6_test.go
rainest: The proxy shouldn't have any configuration at this point; the test doesn't add any and KTF tests run in isolation. If it does somehow appear, something strange has occurred.

PR #755, opened 2023-07-24
refactor: introduce reusable test helpers

programmer04:
Currently, only Kong Enterprise deployed with Postgres is covered with tests. Refactor by extracting common pieces to make it easy to add a new test for DBLess mode.

Moreover, add a check that the workspace (enterprise feature) has been indeed created in a Kong Gateway (checking directly in API since this test checks to create Kong Gateway that runs in enterprise mode.

In the next separate PR support for DBLess will be tested.

### DIFF
--- a/internal/test/assert.go
+++ b/internal/test/assert.go
@@ -4,56 +4,102 @@ import (
 	"bytes"
 	"io"
 	"net/http"
+	"strings"
 	"testing"
 	"time"
 
+	gokong "github.com/kong/go-kong/kong"
 	"github.com/stretchr/testify/require"
 )
 
-type httpResponseOption struct {
-	bodyChecker       func(*testing.T, []byte) bool
-	statusCodeChecker func(t *testing.T, reqURL string, statusCode int, body []byte) bool
+type configurationOption struct {
+	responseChecker []func(*testing.T, *http.Response) bool
+	waitFor         time.Duration
+	interval        time.Duration
 }
 
-type HTTPResponseOpt func(*httpResponseOption)
+type ConfigurationOpt func(*configurationOption)
 
-func WithBodyChecker(bodyChecker func(*testing.T, []byte) bool) HTTPResponseOpt {
-	return func(opts *httpResponseOption) {
-		opts.bodyChecker = bodyChecker
+func WithWaitFor(waitFor time.Duration) ConfigurationOpt {
+	return func(opts *configurationOption) {
+		opts.waitFor = waitFor
 	}
 }
 
-func WithBodyContains(expected string) HTTPResponseOpt {
-	return func(opts *httpResponseOption) {
-		opts.bodyChecker = func(t *testing.T, body []byte) bool {
-			t.Logf("check expected content %s of the response body", expected)
-			return bytes.Contains(body, []byte(expected))
-		}
+func WithInterval(interval time.Duration) ConfigurationOpt {
+	return func(opts *configurationOption) {
+		opts.interval = interval
 	}
 }
 
-func WithStatusCode(expected int) HTTPResponseOpt {
-	return func(opts *httpResponseOption) {
-		opts.statusCodeChecker = func(t *testing.T, reqURL string, statusCode int, body []byte) bool {
-			if statusCode != expected {
-				t.Logf("WARNING: unexpected response %s: %d with body: %s", reqURL, statusCode, body)
+func WithResponseChecker(bodyChecker func(*testing.T, *http.Response) bool) ConfigurationOpt {
+	return func(opts *configurationOption) {
+		opts.responseChecker = append(opts.responseChecker, bodyChecker)
+	}
+}
+
+func WithBodyContains(expected string) ConfigurationOpt {
+	return WithResponseChecker(
+		func(t *testing.T, resp *http.Response) bool {
+			body, err := io.ReadAll(resp.Body)
+			if err != nil {
+				t.Logf("WARNING: error cannot read response body %s: %v", resp.Request.URL, err)
 				return false
 			}
-			t.Logf("expected status code %d of the response received", statusCode)
+			if !bytes.Contains(body, []byte(expected)) {
+				t.Logf("WARNING: unexpected content of response body %s: %s", resp.Request.URL, body)
+				return false
+			}
+			t.Logf("expected content of the response body received")
 			return true
-		}
-	}
+		},
+	)
 }
 
-// EventuallyExpectedStatusCodeAndBody is a helper function that retries the request until
+func WithStatusCode(expected int) ConfigurationOpt {
+	return WithResponseChecker(
+		func(t *testing.T, resp *http.Response) bool {
+			if resp.StatusCode != expected {
+				t.Logf("WARNING: unexpected status code %d, expected %d", resp.StatusCode, expected)
+				return false
+			}
+			t.Logf("expected status code received")
+			return true
+		},
+	)
+}
+
+func WithEnterpriseHeader() ConfigurationOpt {
+	return WithResponseChecker(
+		func(t *testing.T, resp *http.Response) bool {
+			version, _ := strings.CutPrefix(resp.Header.Get("Server"), "kong/")
+			v, err := gokong.NewVersion(version)
+			if err != nil {
+				t.Logf("WARNING: error while parsing admin api version %s: %v", version, err)
+				return false
+			}
+			t.Logf("admin api version %s", v)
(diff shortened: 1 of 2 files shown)


test/integration/enterprise_test.go
```diff
@@ -67,77 +73,127 @@ func TestKongEnterprisePostgres(t *testing.T) {
...
+
+func prepareKongEnterpriseLicense(t *testing.T) string {
+	t.Helper()
+	t.Log("preparing kong enterprise license")
+	licenseJSON, err := kong.GetLicenseJSONFromEnv()
+	require.NoError(t, err)
+	return licenseJSON
+}
+
+func eventuallyExpectedStatusCodeAndBody(
```
pmalek, line 170: Can you consider putting this in a place where it can be reused (not only by integration tests) e.g. in https://github.com/Kong/kubernetes-ingress-controller/blob/dcd58c38c4b53689a863e6a0e67d4684ab37df13/test/internal/helpers/http.go and exporting it?

programmer04: Good idea, I'm doing it

programmer04: But, you're referencing the KIC repo, and here is KTF. The place where can I put it (and it can be reused by everything that imports KTF) is [pkg/utils](https://github.com/Kong/kubernetes-testing-framework/tree/c043d71850708036bf205f0d0553b5b5661bbb32/pkg/utils). WDYT?

pmalek: No `utils` please 🙃 We can introduce a new package which will aggregate testing types and functions. Could be placed under `pkg/test/` (this project is named `*-testing-framework` after all).

Ideally we'd think twice about the API so that it could be reused and not be too specific. E.g. introduce option parameters that could be applied on the HTTP response etc.

programmer04: Yeah, but now I'm wondering if helpers tight to the specific testing library should be part of the public API of the package...

pmalek: Do you have any other suggestions? This can be reused in many places e.g. KGO, KIC etc. We could theoretically use a separate repo for those but I don't see a reason at this stage.

Perhaps we would go this route if several of those "test helpers" were moved (from e.g. KIC) to said repo and imported in KIC and other projects.

programmer04: Nothing concrete. I have the same feeling as you and see the benefit. I think we should discuss it with the team and go through all repos and check what we have already and where/how we want to put it. And for this PR, leaving it as is for now

pmalek: Ok agreed. Let's create an issue: can be in e.g. team-k8s repo and discuss this on next team sync.

One point that I'd prefer to resolve here still is the API of this: the interface requires the user to 
- provide a `nil` argument when body checker param is unused. This I believe could be improved by aforementioned options pattern.
- keep rewriting the checks that most likely will be reused. We could also provide the option to e.g. check if the version is enterprise (as used above) and reuse that, making the call site that much cleaner.

Last but not least: I'd suggest to put this in a separate package but given the unresolved debate above we can make this an `internal` package for now, so that we can reuse it in KTF but we don't have to worry about breaking users until we expose it (by placing it in a non `internal` package).

WDYT?

programmer04: I provided opts pattern and introduced an internal package. For checking if it's an enterprise after a couple of tries I couldn't come up with a clean API, so I left it as it is. The issue has been created https://github.com/Kong/team-k8s/issues/292

internal/test/assert.go
```diff
@@ -0,0 +1,62 @@
...
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+type checkerOptions struct {
+	bodyChecker func(*testing.T, []byte) bool
+}
+
+type CheckerOption func(*checkerOptions)
```
pmalek, line 16: This type is called `CheckerOption` in a package called `test`. This might not be ideal since it's a bit vague.

How about calling this `HTTPResponseOpt`? + see below suggestion.

internal/test/assert.go
```diff
@@ -0,0 +1,62 @@
...
+	return func(opts *checkerOptions) {
+		opts.bodyChecker = bodyChecker
+	}
+}
+
+// EventuallyExpectedStatusCodeAndBody is a helper function that retries the request until
+// it gets the expected status. It also checks the body of the response if a body checker
+// is provided. Default is to retry for 1 minute with 1 second interval.
+func EventuallyExpectedStatusCodeAndBody(
+	t *testing.T, httpClient *http.Client, req *http.Request, expectedStatusCode int, opts ...CheckerOption,
```
pmalek, line 28: Instead doing this, how about we make this a generic `http.Response` option + add explicit note about only one option to be able to consume the body. WDYT about that?

test/integration/enterprise_test.go
```diff
@@ -67,77 +73,100 @@ func TestKongEnterprisePostgres(t *testing.T) {
...
-			if err != nil {
+	if adminPassword != "" {
+		decorateRequestWithAdminPassword(t, req, adminPassword)
+	} else {
+		t.Log("skipping setting the admin api password (Kong-Admin-Token header)")
+	}
+	t.Log("verifying the admin api version is enterprise")
+	httpClient := &http.Client{Timeout: time.Second * 10}
+	test.EventuallyExpectedStatusCodeAndBody(t, httpClient, req, http.StatusOK, test.WithBodyChecker(
+		func(t *testing.T, body []byte) bool {
```
pmalek, line 86: This whole check / option I believe could be made into a named option like e.g. `WithBodyHasKongEnterpriseVersion`. WDYT?

test/integration/enterprise_test.go
```diff
@@ -67,77 +73,100 @@ func TestKongEnterprisePostgres(t *testing.T) {
...
+		nil,
+	)
 	require.NoError(t, err)
-	req.Header.Set("kong-admin-token", adminPassword)
+	test.EventuallyExpectedStatusCodeAndBody(t, httpClient, req, http.StatusOK, test.WithBodyChecker(
+		func(t *testing.T, body []byte) bool {
+			const expectedBody = "<title>httpbin.org</title>"
+			t.Logf("check expected content %s of the response body", expectedBody)
+			return bytes.Contains(body, []byte(expectedBody))
+		},
```
pmalek, line 126: Ditto about making this a named option, like e.g. `WithBodyContains`.

internal/test/assert.go
pmalek: + add an option checking status code.

test/integration/enterprise_test.go
programmer04: But you need a combination of the right request and function for checking, here we explicitly pass a request

programmer04: That `WithBodyHasKongEnterpriseVersion()` will work only when `req` will contain a valid URL and method. Another approach is to check it based on the header returned by Gateway, but idk if the format is the same as in current approach

pmalek: I don't mind using body or response header: whatever works. I'm just concerned about the interface here.

I don't see a problem with the request not containing a proper URL or a method. It will just failed on `http.Do()` call right? Before the option kicks in.

internal/test/assert.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+type configurationOption struct {
+	responseChecker []func(*testing.T, *http.Response) bool
+	waitFor         time.Duration
+	interval        time.Duration
+}
+
+type ConfigurationOpt func(*configurationOption)
```
pmalek, line 21: Nit: this name might still be a bit vague for the user. How about something along the lines of `HTTPResponseExpectOpt`?

internal/test/assert.go
```diff
@@ -0,0 +1,130 @@
...
+			}
+			return true
+		},
+	)
+}
+
+// EventuallyExpectedResponse is a helper function that retries the request until
+// it gets the expected response. For setting expected status code use WithStatusCode
+// option (otherwise it's not checked). For checking content of body only one has to be
+// passed (body can be read only one time).
```
pmalek, line 94:
```suggestion
// EventuallyExpectedResponse is a helper function that issues the provided request
// until it gets the expected response or the timeout (default: 1 minute) is reached.
// For assertions about the received response one can provide options like WithStatusCode or WithBodyContains.
// NOTE: only one option checking the body can be provided since body's reader can only
// be read once.
```

pmalek (approved):
:shipit:

PR #757, opened 2023-07-25
fix: allow to deploy Kong Gateway EE in DBLess mode

programmer04:
This PR fixes KTF to allow deployment of Kong Gateway EE in DBLess mode. New introduced test `TestKongEnterpriseDBLess` checks it. The adjustment has to be made for tests for the enterprise version. Instead of using `/workspaces` that silently doesn't work in DBLess mode when trying to configure them - nothing is created (will be reported to the upstream, but it's rather considered as normal, old, expected behavior) `/consumer_groups` are used. Consumer Groups work in the same way in both versions.

### DIFF
--- a/test/integration/enterprise_test.go
+++ b/test/integration/enterprise_test.go
@@ -22,7 +22,6 @@ import (
 
 func TestKongEnterprisePostgres(t *testing.T) {
 	SkipEnterpriseTestIfNoEnv(t)
-	t.Parallel()
 
 	licenseJSON := prepareKongEnterpriseLicense(t)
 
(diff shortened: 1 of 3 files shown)


pmalek:
`TestKongEnterprisePostgres` fails :(

programmer04:
flake 🥲 
```
 Error: UPGRADE FAILED: failed to create resource: Service "ingress-controller-kong-proxy" is invalid: spec.ports[0].nodePort: Invalid value: 30080: provided port is already allocated
```
 @pmalek

test/integration/enterprise_test.go
```diff
@@ -97,26 +110,29 @@ func deployAndTestKongEnterprise(t *testing.T, kongAddon *kongaddon.Addon, admin
...
+		t.Log("verifying enterprise consumer groups API functionality using /config (works only for dblessmode)")
+		req, err = http.NewRequestWithContext(
+			ctx, http.MethodPost, adminURL.JoinPath("/config").String(),
+			strings.NewReader(fmt.Sprintf(`{"_format_version": "3.0", "consumer_groups": [{"name": "%s"}]}`, consumerGroupToCreate)),
+		)
 	}
 	req.Header.Set("Content-Type", "application/json")
 	test.EventuallyExpectResponse(t, httpClient, req, test.WithStatusCode(http.StatusCreated))
 
 	t.Log("verifying that the workspace was indeed created")
```
czeslavo, line 132:
```suggestion
	t.Log("verifying that the consumer group was indeed created")
```

czeslavo (approved):

czeslavo (approved):

PR #762, opened 2023-07-26
feat: add ArgoCD and ArgoCD Kong addons

rainest:
Add an addon that installs ArgoCD core.

Add an addon that installs the Kong Helm chart using ArgoCD.

This is currently quite bare bones as far as Kong options. It's intended as a base viable install before figuring out what else we'll need for tests in service of https://github.com/Kong/charts/issues/763.

My expectation is that most users will continue to use the existing Kong addon, with the ArgoCD variant being primarily for our own chart compatibility tests.

We could go with a lighter weight alternative that only adds the ArgoCD plugin and expects users to deploy Application/AppProject manifests outside KTF, but having an app as a plugin makes it easier to meaningfully test the ArgoCD plugin. There's also a minor gotcha where applying Applications before ArgoCD becomes ready apparently causes issues deploying them (unsure if there are Application settings I can tune to address this), so being able to explicitly wait on the other plugin in the app plugin's deploy method avoids that.


test/integration/kong_argo_addon_test.go
```diff
@@ -0,0 +1,55 @@
...
+	t.Logf("setting up the environment cleanup for environment %s and cluster %s", env.Name(), env.Cluster().Name())
+	defer func() {
+		t.Logf("cleaning up environment %s and cluster %s", env.Name(), env.Cluster().Name())
+		require.NoError(t, env.Cleanup(ctx))
+	}()
+
+	t.Log("verifying that addons have been loaded into the environment")
+	require.Len(t, env.Cluster().ListAddons(), 3)
+
+	t.Log("verifying that the kong deployment is using custom images")
```
pmalek, line 51: What is meant here by `kong is using custom images`? Either I missed it or no "custom images" have been mentioned in here 🤔 

The code below actually only checks if the kong addon is ready

pkg/clusters/addons/argocd/addon.go
```diff
@@ -0,0 +1,176 @@
...
+const (
+	// AddonName indicates the unique name of this addon.
+	AddonName clusters.AddonName = "argocd"
+
+	// DefaultNamespace indicates the default namespace this addon will be deployed to.
+	DefaultNamespace = "argocd"
+
+	canaryDeployment = "argocd-repo-server"
+
+	manifestURL = "https://raw.githubusercontent.com/argoproj/argo-cd/stable/manifests/core-install.yaml"
```
czeslavo, line 30: As `stable` sounds like a reasonable default, I suggest we also allow using a pinned version avoid unexpected changes kicking in with no further ado. I'd expect us to pin this version in any of our tests.
```suggestion
	manifestURLTemplate = "https://raw.githubusercontent.com/argoproj/argo-cd/%s/manifests/core-install.yaml"
	defaultTag = "stable"
```
The builder could have `WithVersion` method to allow using versions different than `stable`.

pmalek: 👍 for using a pinned version. We already had issues with a floating version pointer (like a branch) in the past.

test/integration/kong_argo_addon_test.go
czeslavo, line 51: We do not verify the Kong deployment uses a custom image, are we?

pkg/clusters/addons/kongargo/addon.go
```diff
@@ -0,0 +1,227 @@
...
+	ns := corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: a.namespace}}
+	if _, err := cluster.Client().CoreV1().Namespaces().Create(ctx, &ns, metav1.CreateOptions{}); err != nil {
+		if !errors.IsAlreadyExists(err) {
+			return err
+		}
+	}
+
+	argo, err := getArgo(cluster)
+	if err != nil {
+		return err
```
czeslavo, line 83: I think we could make use of error wrapping here to provide better context in case of failures (it applies to all places in this PR where we return err without wrapping it).
```suggestion
		return fmt.Errorf("failed to get ArgoCD addon: %w", err)
```

test/integration/kong_argo_addon_test.go
rainest: Garbage left over from a copied test skeleton; updated.

pkg/clusters/addons/argocd/addon.go
rainest: Borrowing the configurable approach from Istio, minus the lookup (since Argo provides their own "latest stable". You get a floating tag by default, but can `WithVersion()` to pin one as a KTF client, as a balance between stability and not needing to do a new release to get updates.

test/integration/kong_argo_addon_test.go
```diff
@@ -0,0 +1,57 @@
...
+	}()
+
+	t.Log("verifying that addons have been loaded into the environment")
+	require.Len(t, env.Cluster().ListAddons(), 3)
+
+	t.Log("verifying that the ArgoCD creates a Kong deployment and it reaches the ready state")
+	require.Eventually(t, func() bool {
+		_, ready, err := kong.Ready(ctx, env.Cluster())
+		return err == nil && ready
+	}, time.Minute*3, time.Second)
```
czeslavo, line 56: I think we should have test code that verifies the desired Kubernetes Deployment really gets created and becomes ready. Now we rely on the code that we're testing to do that. If `kong.Ready` implementation changes, this test won't catch it.

rainest: We can kinda copy the current contents into the test, but it seems a bit silly to do so. We should trust that plugin implementations provide a meaningful Ready function that does serve as a valid oracle for whether the plugin assets are ready for use. We control that code and should reject changes that violate that contract.

Running the check in the test would require specifying the Deployment name, which isn't normally exposed by the plugin and may itself change, so I'd say leave this as-is.

rainest:
TestEnvironmentWithRegistryAddon flaked, but the error is inscrutable: it dumps the entire Deployment and then says "should have 0 items, but has 1". Damnit Kubernetes, that means nothing.

test/integration/kong_argo_addon_test.go
czeslavo: > We control that code and should reject changes that violate that contract.

Sounds fair 👍

czeslavo (approved):

pkg/clusters/addons/argocd/addon.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func New() clusters.Addon {
+	return &Addon{}
+}
+
+// importing the argocd package requires replacing a bunch of Kubernetes libs for reasons:
+// https://argo-cd.readthedocs.io/en/stable/user-guide/import/
+// unfortunately, the library versions are ancient and not compatible with what KTF wants, so we need to use
+// unstructured types
```
pmalek, line 50: 😿

PR #770, opened 2023-08-08
fix: set udpProxy.type to proxyServiceType specified

### LINKED ISSUE #540: Add option to configure kong UDP service type with `WithProxyServiceType()` or similar

### Problem Statement

https://github.com/Kong/kubernetes-ingress-controller/pull/3325 introduced udp service which is deployed by default.

This should be configured via `WithProxyServiceType()` (or separate builder func?) similarly as its being done for the "regular"/tcp service.

### Proposed Solution

- configure udp service type via a helm arg (which might need to be introduced in the chart) similarly as it's for the tcp service: https://github.com/Kong/kubernetes-testing-framework/blob/5b3d5ca2fbb2b4ec55b96ac3d026baa903d8980c/pkg/clusters/addons/kong/addon.go#L295

### Additional information

_No response_

### Acceptance Criteria

_No response_

randmonkey:
Fix: set `udpProxy.type` to specified proxy service type, because `udpProxy` service is enabled by default and has type `LoadBalancer` by default. By setting this we can set `proxyServiceType` to `ClusterIP` to get rid of dependency on `metallb` in unnecessary test cases. Fixes #540.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Support specifying namespace of Kong addon to deploy Kong in certain
   namespace.
   [#766](https://github.com/Kong/kubernetes-testing-framework/pull/766)
+- Fix setting UDP proxy service type in Kong addon.
+  [#770](https://github.com/Kong/kubernetes-testing-framework/pull/770)
 
 ## v0.34.0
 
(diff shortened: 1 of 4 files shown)


pmalek (commented):
We might want to add a comment that https://github.com/Kong/kubernetes-testing-framework/blob/5b3d5ca2fbb2b4ec55b96ac3d026baa903d8980c/pkg/clusters/addons/kong/builder.go#L158-L159 will also set the service type for UDP service.

randmonkey:
> We might want to add a comment that
> 
> https://github.com/Kong/kubernetes-testing-framework/blob/5b3d5ca2fbb2b4ec55b96ac3d026baa903d8980c/pkg/clusters/addons/kong/builder.go#L158-L159
> 
> will also set the service type for UDP service.

done.

pmalek (approved):

PR #800, opened 2023-09-05
chore(go): introduce usage of Go toolchain

programmer04:
**What this PR does / why we need it**:

Embrace [native toolchain management](https://go.dev/blog/toolchain) introduced in Go 1.21.0. Moreover do not rely on the hardcoded Go version in GH workflows that has to be updated manually, use the version defined in `go.mod` instead, [read more].

To conclude when a new version of Go is released only update in `go.mod` to have the newest (and the same) version everywhere used (even locally).

**Special notes for your reviewer:**

Probably in the future usage of `actions/setup-go@v4` to set version based on `go.mod` can be ditched too, because the Go toolchain ensures the right version automatically. For now, it's required because GH runners have an older version of Go that does not support toolchain features.


test/e2e/gke_cluster_test.go
```diff
@@ -60,7 +60,7 @@ func testGKECluster(t *testing.T, createSubnet bool) {
 
 	t.Logf("configuring the GKE cluster PROJECT=(%s) LOCATION=(%s)", gkeProject, gkeLocation)
 	builder := gke.NewBuilder([]byte(gkeCreds), gkeProject, gkeLocation)
-	builder.WithClusterMinorVersion(1, 23)
+	builder.WithClusterMinorVersion(1, 24)
```
pmalek, line 63: This should be a separate change IMHO.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/kong/kubernetes-testing-framework
 
-go 1.19
+go 1.21
+
+toolchain go1.21.0
```
pmalek, line 5: While I understand the motivation I also see this as an end user facing change because this will bump  the minimum required version to compile ktf (and modules that require it) 🤔 

OTOH though we're still pre v1 with ktf so I guess we should be fine.

test/e2e/gke_cluster_test.go
programmer04: Separate PR https://github.com/Kong/kubernetes-testing-framework/pull/801

pmalek:
Please rebase to include changes from https://github.com/Kong/kubernetes-testing-framework/pull/801

test/e2e/gke_cluster_test.go
pmalek: I wonder why it doesn't show a conflict given that #801 got already merged 🤔

pmalek (approved):

PR #834, opened 2023-10-10
fix(metallb): remove conflicting IPAddressPools when metallb admission webhook rejects ktf's pool

### LINKED ISSUE #832: `metallb` addon fails to deploy when already existing `IPAddressPool` conflicts with the one to be deployed
### Problem statement

When trying to deploy the `metallb` addon (or another addon that depend on it) and there already exists an `IPAddressPool` which conflicts with the one to be deployed then the following error is blocking the deployment:

```
Message: "admission webhook \"ipaddresspoolvalidationwebhook.metallb.io\" denied the request: CIDR \"172.18.0.100/30\" in pool \"ktf-pool\" overlaps with already defined CIDR \"172.18.0.100/30\"",
Reason: "CIDR \"172.18.0.100/30\" in pool \"ktf-pool\" overlaps with already defined CIDR \"172.18.0.100/30\"",
Details: *k8s.io/apimachinery/pkg/apis/meta/v1.StatusDetails nil,
Code: 403
```

type of the error: https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#StatusError

relevant piece of code: https://github.com/Kong/kubernetes-testing-framework/blob/05a2a38c579797177e1614b60d730b2983b80155/pkg/clusters/addons/metallb/metallb.go#L225-L254

pmalek:
Fixes #832

### DIFF
--- a/pkg/clusters/addons/metallb/metallb.go
+++ b/pkg/clusters/addons/metallb/metallb.go
@@ -13,7 +13,7 @@ import (
 
 	"go4.org/netipx"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -43,6 +43,8 @@ const (
 
 	addressPoolName     = "ktf-pool"
 	l2AdvertisementName = "ktf-empty"
+
+	tickTime = 100 * time.Millisecond
 )
 
 var (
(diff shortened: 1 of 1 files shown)


rainest (commented):
Should we not just have an option to disable pool creation? Deleting whatever's there without knowing what it is seems a bit much.

pmalek:
> Should we not just have an option to disable pool creation? Deleting whatever's there without knowing what it is seems a bit much.

I see your point OTOH if with a configuration option you wouldn't be able to back off from a failed e.g. test run using ktf. With such a mechanism it is able "to recover".

I'm open to suggestions on this one.

rainest:
There aren't any situations where we expect other IPAddressPools to appear without the user expecting them to be there, right? In the case that raised this, it's because the user both enabled the metallb addon and provided their own IPAddressPool. An option to disable addon pool creation enables that bring-your-own pool approach.

pmalek:
https://github.com/Kong/kubernetes-testing-framework/pull/835 introduced disabling `IPAddressPool`. This will have to be included in e.g. KIC through e.g. an env flag when running tests.

PR #844, opened 2023-10-20
feat(kong): allow setting Node ports

pmalek:
Add possibility to specify NodePorts for kong's admin API and HTTP port.

This is required for https://github.com/Kong/kubernetes-ingress-controller/pull/4823 to prevent tests colliding when deploying multiple kong addons to the same cluster.

### DIFF
--- a/pkg/clusters/addons/kong/addon.go
+++ b/pkg/clusters/addons/kong/addon.go
@@ -95,6 +95,10 @@ type Addon struct {
 	proxyEnvVars                      map[string]string
 	proxyReadinessProbePath           string
 
+	// Node ports
+	httpNodePort  int
+	adminNodePort int
+
 	// proxy server enterprise mode configuration options
 	proxyEnterpriseEnabled            bool
 	proxyEnterpriseSuperAdminPassword string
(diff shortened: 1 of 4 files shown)


pkg/clusters/addons/kong/addon.go
```diff
@@ -372,6 +376,14 @@ func (a *Addon) Deploy(ctx context.Context, cluster clusters.Cluster) error {
...
 	args = append(args, a.deployArgs...)
 	args = append(args, defaults()...)
+
+	if a.httpNodePort > 0 {
+		args = append(args, "--set", fmt.Sprintf("proxy.http.nodePort=%d", a.httpNodePort))
+	}
+	if a.adminNodePort > 0 {
+		args = append(args, "--set", fmt.Sprintf("admin.http.nodePort=%d", a.adminNodePort))
+	}
+
```
shaneutt, line 386: :+1: 

(on a side note, it's been a while since I have looked at this code, I forgot how big these Deploy methods got! Not something we need to solve here though)

pkg/clusters/addons/kong/vars.go
```diff
@@ -53,4 +53,8 @@ const (
 	// DefaultProxyNodePort indicates the default NodePort that will be used for
 	// the proxy when applicable.
 	DefaultProxyNodePort = 30080
+
+	// DefaultAdminNodePort indicates the default NodePort that will be used for
+	// the admin API when applicable.
+	DefaultAdminNodePort = 32080
```
shaneutt, line 59: We added this, but I'm not seeing where we're using this? Is this just for caller reference, or am I missing something?

pmalek: Indeed we're not using this 🤔 Has to be an artifact of trying to be consistent with defining that as it's done for proxy port. I'm open to both leaving or removing this.

shaneutt: I think it's fine, was just making sure we weren't _mistakenly_ not using it.

shaneutt (approved):

PR #949, opened 2024-02-02
feat: use kuma version when installing it

czeslavo:
In Kuma addon builder we allow setting a version, but we don't use it when calling `helm install`.

### DIFF
--- a/pkg/clusters/addons/kuma/addon.go
+++ b/pkg/clusters/addons/kuma/addon.go
@@ -61,7 +61,7 @@ func (a *Addon) Namespace() string {
 	return Namespace
 }
 
-// Version returns the version of the Kuma Helm chart that will be deployed by the addon.
+// Version returns the version of the Kuma Helm chart deployed by the addon.
 // If the version is not set, the second return value will be false and the latest local
 // chart version will be used.
 func (a *Addon) Version() (v semver.Version, ok bool) {
--- a/pkg/clusters/addons/kuma/addon.go
+++ b/pkg/clusters/addons/kuma/addon.go
@@ -45,7 +45,7 @@ type Addon struct {
 	name   string
 	logger *logrus.Logger
 
-	version semver.Version
+	version *semver.Version
 
 	mtlsEnabled bool
 }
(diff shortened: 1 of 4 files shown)


pkg/clusters/addons/kuma/addon.go
```diff
@@ -63,7 +63,10 @@ func (a *Addon) Namespace() string {
 
 // Version indicates the Kuma version for this addon.
 func (a *Addon) Version() semver.Version {
-	return a.version
+	if a.version == nil {
+		return semver.Version{}
+	}
+	return *a.version
```
shaneutt, line 69: The downside of this change is that you'll get a `v0.0.0` even if you didn't set a version, so the caller would need to guard for this.

It's been a while since I dug around in this code, but couldn't we get away with changing the signature for this function instead?

```suggestion
func (a *Addon) Version() (semver.Version, error) {
```

And return an error when not set? :thinking:

pkg/clusters/addons/kuma/addon.go
```diff
@@ -144,6 +147,10 @@ func (a *Addon) Deploy(ctx context.Context, cluster clusters.Cluster) error {
 	// if the dbmode is postgres, set several related values
 	args := []string{"--kubeconfig", kubeconfig.Name(), "install", DefaultReleaseName, "kuma/kuma"}
 
+	if a.version != nil {
+		args = append(args, "--version", a.version.String())
+	}
+
```
shaneutt, line 153: Haha, whoops! That's been missing for quite a while hasn't it...

pkg/clusters/addons/kuma/addon.go
czeslavo: Yup. Technically that is a breaking change, but given we're still before v1 we have an excuse I guess. 👍 4963e7b

pmalek (commented):
One nit, I'd have is to include this change in the changelog. Other than that looks good.

programmer04:
When your PR will be merged and this https://github.com/Kong/kubernetes-testing-framework/pull/946 too I think we can release a new version

programmer04 (approved):

programmer04 (approved):
👍

programmer04 (approved):

PR #1017, opened 2024-03-27
Backport Kuma updates to 0.39

rainest:
Backport https://github.com/Kong/kubernetes-testing-framework/commit/fe761fa03cc425dbc3c76bcd35246b5fc41b9ca7 and https://github.com/Kong/kubernetes-testing-framework/commit/78164723d50b1012015d938c9cf4dc7720219c3e to 0.39 and release 0.39.2.

Update CI to test on PRs to release branches. Update license handling for this release branch.

---

KIC 2.12 is able to pass E2E with this in place: https://github.com/Kong/kubernetes-ingress-controller/pull/5760#issuecomment-2026047921

---

This technically breaks semver, but we don't really have anywhere else to stuff it. Maybe you can call the addon not supporting version args and not supporting 2.6 "bugs" for the sake of discussion :shrug: 

Updating KIC 2.x to use the latest KTF causes [all sorts of conflicts](https://github.com/Kong/kubernetes-ingress-controller/pull/5759) due to dependency hell with packages that both KTF and KIC import. We don't have a compelling need to actually update those in 2.x, so trying to make a backported KTF version that can run Kuma tests without pulling updates is much simpler.

@randmonkey can you confirm the exact nature of the Kuma failures? Was it just that it was trying to use latest without the 2.6 fix? IDK if we'd also need to pin the Kuma version in the 2.x KIC tests after this.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -3,12 +3,12 @@ name: lint
 on:
   pull_request:
     branches:
-    - '*'
+    - '**'
   push:
     branches:
     - 'main'
     tags:
-    - '*'
+    - '**'
   workflow_dispatch: {}
 
 jobs:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -10,6 +10,7 @@ on:
   pull_request:
     branches:
     - 'main'
+    - 'release/[0-9]+.[0-9]+.x'
   push:
     branches:
     - 'main'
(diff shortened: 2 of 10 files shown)


randmonkey:
According to @czeslavo discovered, After we add the `allowAllTrafficPermission` settings, the test could pass with kuma 2.6.x. But we still need to configure the version in the kuma addon, otherwise KTF cannot recognize the version thus it will not add the settings.

randmonkey (dismissed):

czeslavo (approved):

PR #1106, opened 2024-06-17
chore: extend e2e tests' timeout

czeslavo:
Drop the e2e tests job's timeout set from organization's vars as it's not enough: https://github.com/Kong/kubernetes-testing-framework/actions/runs/9550375317/job/26324041269

### DIFF
--- a/.github/workflows/release-testing.yaml
+++ b/.github/workflows/release-testing.yaml
@@ -86,7 +86,7 @@ jobs:
         TEST_RUN: ${{ matrix.test }}
 
   e2e-tests:
-    timeout-minutes: ${{ fromJSON(vars.GHA_DEFAULT_TIMEOUT) }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release-testing.yaml
```diff
@@ -86,7 +86,6 @@ jobs:
         TEST_RUN: ${{ matrix.test }}
 
   e2e-tests:
-    timeout-minutes: ${{ fromJSON(vars.GHA_DEFAULT_TIMEOUT) }}
```
pmalek, line 89: Can you we set our own timeout instead?

czeslavo: Yup, setting it to 30m as most of the runs oscillated between 10-25m.

pmalek (approved):

PR #1258, opened 2025-01-15
feat: add network policy switch for creating GKE clusters

### LINKED ISSUE #1249: Support turning on network policy when creating GKE clusters

### Problem Statement

In certain test scenarios, we need to enable network policy on the cluster and it's not possible to achieve using current gke cluster `Builder`.

### Proposed Solution

- Add a new function into the [gke Builder](https://github.com/Kong/kubernetes-testing-framework/blob/0e167c0aca2a53a8e92ec8f2ab887760eaf4eb94/pkg/clusters/types/gke/builder.go) and mark the builder to set `NetworkPolicy` field when creating the cluster.

https://github.com/googleapis/google-cloud-go/blob/main/container/apiv1/containerpb/cluster_service.pb.go#L7039-L7040
```
// A Google Kubernetes Engine cluster.
type Cluster struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	// The name of this cluster. The name must be unique within this project
	// and location (e.g. zone or region), and can be up to 40 characters with
	// the following restrictions:
	//
	// * Lowercase letters, numbers, and hyphens only.
	// * Must start with a letter.
	// * Must end with a number or a letter.
	Name string `protobuf:"bytes,1,opt,name=name,proto3" json:"name,omitempty"`

        // ...

	// Configuration options for the NetworkPolicy feature.
	NetworkPolicy *NetworkPolicy `protobuf:"bytes,19,opt,name=network_policy,json=networkPolicy,proto3" json:"network_policy,omitempty"`

        // ...
}
```

### Additional information

_No response_

### Acceptance Criteria

- [ ] As a user, I can enable network policy when I ceate the environment using GKE cluster type, so that my cluster will have calico enabled by default once the cluster is ready.

jijiechen:
fixes #1249

This PR also use the new `NodePools` field to replace the deprecated `NodeConfig`.

The implementation has been tested manually and a CI run is available at https://github.com/kumahq/kuma-smoke/actions/runs/12667973965/job/35302508845


pkg/clusters/types/gke/builder.go
```diff
@@ -205,6 +217,12 @@ func (b *Builder) Build(ctx context.Context) (clusters.Cluster, error) {
 			Channel: channel,
 		}
 	}
+	if b.enableNetworkPolicy {
+		pbcluster.NetworkPolicy = &containerpb.NetworkPolicy{
+			Enabled:  true,
+			Provider: containerpb.NetworkPolicy_CALICO,
```
pmalek, line 223: While currently it seems that only CALICO is allowed here (ref: https://pkg.go.dev/cloud.google.com/go/container@v1.42.0/apiv1/containerpb#NetworkPolicy_Provider) how about making the flag not a bool (and set Calico as the implementation of choice) but rather make the flag accept a string and validate it against the allowed set of implementations?

This way in the future we'll be able to add more options to choose as the implementation. WDYT?

jijiechen: Well, I guess they'll never support a second network policy provider in standard clusters, so by using a boolean can simplify the API here.  
They've created Dataplane v2 which has network policy support built-in, and in auto-pilot clusters, they use Dataplane v2 by default. So I guess their standard clusters are going to be sort of "legacy" in the near future.

jijiechen: Source: https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy#calico_migrate

pmalek: Fair enough. Thanks for the context and the link 👍 

I just wanted to make sure that we're not locking ourselves in for the future. 👍

pmalek (approved):
Do you mind adding a changelog entry in https://github.com/Kong/kubernetes-testing-framework/blob/main/CHANGELOG.md for this change?

pmalek (approved):

pmalek:
@jijiechen with #1263 merged we should be able to proceed with this PR when you rebase on latest `main` tip.

jijiechen:
Hi thanks for noting this. I've just merged master.

jijiechen:
Some of the CI jobs are failing because of rate limiting in CodeCov tasks

pmalek:
@jijiechen Sorry for this. The issues with codecov should be resolved now as https://github.com/Kong/kubernetes-testing-framework/pull/1274 has merged. Please rebase on `main`'s tip.

Issue #285, opened 2022-06-07
AWS EKS cluster support


shaneutt:
### Problem Statement

Historically we've only had support for `kind` clusters and GKE, but with the upcoming [Gateway Operator](https://github.com/kong/gateway-operator) there's need for EKS clusters in CI.

Drivers for this FR:
- Currently it's the best known way to test host-based load balancer services.
- Popularity of AWS meaning that we should test on AWS on a regular basis.

### Acceptance Criteria

- [ ] as a user of the Go library I can create EKS based `cluster.Clusters` and testing environments

PR #1259, opened 2025-01-16
feat(eks): add eks cluster builder

jijiechen:
closes https://github.com/Kong/kubernetes-testing-framework/issues/285

### Features

This PR provides the following features:
* a cluster `Builder` for creating EKS clusters on AWS
* a `Cluster` representing the cluster created and provides the functionality of cleaning up the cluster and all underlying resources created
* a `NewFromExisting` helper function to retrieve kubeconfig from an existing cluster

Above features are tested manually and a CI run can be found here:
https://github.com/kumahq/kuma-smoke/actions/runs/12785877388/job/35641958835

### Implementation

Unlike `eksctl` which creates the cluster using AWS CloudFormation, this PR creates underlying AWS resources directly using [the Golang SDK](https://github.com/aws/aws-sdk-go-v2) provided by AWS. 

* The entrypoint for creating a new cluster is the function `eks.aws_operations.CreateEKSClusterAll`

* The entrypoint for cleaning up a cluster is the function `eks.aws_operations.DeleteEKSClusterAll`

Reused the logic of `eksctl` to generate `userdata` for bootstraping the nodes.

### Limitations

1. only IPv4 networking is supported as of now (this will be resolved by a new feature plan)
2. nodes created for now are not "managed" nodes, they are backed by launch templates (this will be resolved by a new feature plan)
3. no load balancer integration was tested (this will be resolved by a new feature/test plan)
4. only tested using a fedarated user and did not test IAM users with programmatic access as of now (technically, it should work for free as we are using the official SDK to read credentials, and I will test as soon as I get a user with programmatic access)
5. does not provide any retry mechanism when calling AWS APIs, this could lead to unstable API calls and in-complete cluster creation/cleanup.
6. cluster cleanup takes more time than the gke version implementation, because AWS does not provide a "fire-and-forget" API to use for cleaning up all the resources involved.
7. due to AWS platform characteristics, a cluster creation attempt takes ~15mins and a deletion takes ~10mins.
8. due to AWS EKS limitation, the cluster client kube config has a maximum of 900s (15mins). So for long running tests, we need to export the client config peridlically to refresh the validity. 

### Usage

```go
	t.Log("configuring EKS cloud environment for tests")
	require.NotEmpty(t, EKSAccessKeyId, "%s not set", eks.EnvAccessKeyId)
	require.NotEmpty(t, EKSAccessKey, "%s not set", eks.EnvAccessKey)
	require.NotEmpty(t, EKSRegion, "%s not set", eks.EnvRegion)

	t.Logf("configuring the EKS cluster KEY_ID=(%s) REGION=(%s)", EKSAccessKeyId, EKSRegion)
	builder := eks.NewBuilder()
	builder.WithClusterVersion(fmt.Sprintf("%d.%d.0", EKSVersionMajor, EKSVersionMinor))

	t.Logf("building cluster %s (this can take some time)", builder.Name)
	cluster, err := builder.Build(ctx)
	require.NoError(t, err)

	t.Logf("setting up cleanup for cluster %s", cluster.Name())
	t.Cleanup(func() {
		t.Logf("running cluster cleanup for %s", cluster.Name())
		// don't use test ctx as it may be cancelled already
		assert.NoError(t, cluster.Cleanup(context.Background()))
	})

	t.Log("verifying that the cluster can be communicated with")
	version, err := cluster.Client().ServerVersion()
	require.NoError(t, err)
	t.Logf("server version found: %s", version)
```

I'll provide an integration test soon in a new PR.


jijiechen:
this closes https://github.com/Kong/kubernetes-testing-framework/issues/285

pkg/clusters/types/eks/aws-operations/eks.go
```diff
@@ -0,0 +1,479 @@
...
+const (
+	DefaultNodeGroupName     = "default-node-group"
+	DefaultKubernetesSvcCIDR = "172.20.0.0/16"
+	kubernetesTagFormat      = "kubernetes.io/cluster/%s"
+	envKeyNodeSSHKeyName     = "EKS_NODE_SSH_KEY"
+
+	TagNameCreateBy = "ktf_created_by"
+)
+
+func CreateEKSClusterAll(ctx context.Context, cfg aws.Config, clusterName,
```
randmonkey, line 34: It looks like to be a function to create a cluster and all its dependencies. Could we add comments to tell this?

pkg/clusters/types/eks/aws-operations/eks.go
```diff
@@ -0,0 +1,479 @@
...
+	k8sMinorVersion, nodeMachineType string, tags map[string]string) error {
```
randmonkey, line 35: Could we group the parameters with the same type to the same line to make a clearer view on the types of the params?

pkg/clusters/types/eks/aws-operations/eks.go
```diff
@@ -0,0 +1,479 @@
...
+	clusterRoleArn, nodeRoleArn, err := createRoles(ctx, iamClient, clusterName)
+	if err != nil {
+		return errors.Wrap(err, "failed to create IAM roles")
+	}
+	subnetAvZones, err := getAvailabilityZones(ctx, ec2Client, cfg.Region)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get availability zones in region %s", cfg.Region)
+	}
+
+	vpcId, subnetIDs, err := createVPC(ctx, ec2Client, subnetAvZones)
```
randmonkey, line 57: Could it leave dangling resources when error happens after creating a VPC? I think we need to clean them if possible, or add a comment to tell that it may leave resource to clear.

pkg/clusters/types/eks/aws-operations/eks.go
```diff
@@ -0,0 +1,479 @@
...
+
+	err = createNodeGroup(ctx, eksClient, ec2Client, clusterCfg)
+	if err != nil {
+		return errors.Wrapf(err, "failed to create EKS node group for cluster %s", clusterName)
+	}
+
+	return nil
+}
+
+func DeleteEKSClusterAll(ctx context.Context, cfg aws.Config, clusterName string) error {
```
randmonkey, line 117: Ditto here. I think we need to add comments to tell what resources are deleted here.

pkg/clusters/types/eks/aws-operations/eks.go
```diff
@@ -0,0 +1,479 @@
...
+		ng.SSH.PublicKeyName = aws.String(nodeKeyName)
+	}
+
+	return clusterCfg
+}
+
+func waitForClusterActive(ctx context.Context, eksClient *eks.Client, clusterName string) (*types.Cluster, error) {
+	childCtx, cancel := context.WithTimeout(ctx, 10*time.Minute)
+	defer cancel()
+	ticker := time.NewTicker(10 * time.Second)
```
randmonkey, line 223: Is the 10 second ticker too long? Or should we make the wait time and wait tick configurable?

pkg/clusters/types/eks/aws-operations/eks.go
```diff
@@ -0,0 +1,479 @@
...
+	_, err = eksClient.CreateNodegroup(ctx, input)
+	if err != nil {
+		return err
+	}
+
+	return waitForNodeGroupReady(ctx, eksClient, clusterCfg.Metadata.Name, nodeGroup.Name)
+}
+
+func waitForNodeGroupReady(ctx context.Context, eksClient *eks.Client, clusterName, nodeGroupName string) error {
+	childCtx, cancel := context.WithTimeout(ctx, 10*time.Minute)
```
randmonkey, line 299: Ditto here for wait time and wait tick.

pkg/clusters/types/eks/aws-operations/eks.go
```diff
@@ -0,0 +1,479 @@
...
+	}
+
+	err = authorizeNodeGroup(kubeCfg, nodeRoleArn)
+	if err != nil {
+		return errors.Wrapf(err, "failed to authorize node group to access cluster %s", clusterName)
+	}
+
+	amiId, err := resolveAMI(ctx, ec2Client, cfg.Region, k8sMinorVersion, nodeMachineType, eksctlapi.DefaultNodeImageFamily)
+	if err != nil {
+		return errors.Wrap(err, "failed to resolve AMI")
```
tao12345666333, line 94: Can we print the Region or cluster name here?

jijiechen: Cluster name is the environment/cluster-builder name, so it's known from outside.
Region is loaded from the environment variable as well and I'm open to print this. The concern is, this framework tends to not print anything during a cluster building. I'm not sure whether should I print or how should I log since it looks like we don't have a  logging mechanism provided in KTF.

pkg/clusters/types/eks/aws-operations/eks.go
jijiechen: Yes, there can be these orphan resources left when the cluster creation is not complete. I'll provide a better way of cleaning up by enumerate tags in a future PR.

jijiechen: The creation can fail because of different reasons:
* the user does not have permission to perform some of the actions
* API call fails because of network issue

So we'll also need to enhance the experence of these factors.

jijiechen: I'll mark this as  resolved for now as it's not planned in this PR.

pkg/clusters/types/eks/aws-operations/eks.go
jijiechen: Creating a cluster can take up to 10 minutes. So I assume 10 seconds is not too long.

pkg/clusters/types/eks/aws-operations/eks.go
jijiechen: Waiting for a node group to become ready can take about 5 minutes. 
I'll change to 5 seconds here.

